PR #3, opened 2026-03-18
Add "locally cartesian closed" property

varkor:
Note that I have not yet tested these changes locally (I'm currently not a machine on which I can test), but it may be that there is a GitHub Action that tests pull requests. (If not, I should be able to test later today.)


ScriptRaccoon:
> I've updated the pull request. I've added the category `walking composable pair` as well as another property `strongly connected`, which is used to establish local cartesian closure for some categories.

This is great!

I also like that you have added some implications to refactor the properties. This is exactly what CatDat is made for. 🎉

> There are probably some obvious missing implications, but it takes time to establish the properties for every single category, and I don't have more time at the moment.

That's ok :)

> Unfortunately, I'm having trouble getting pnpm to work at the moment, so I still have been unable to test the changes

No problem. There were indeed some minor issues, I have fixed them in feeebc0d83bff0f0c739dddb3893421b935b0be4 (will be squashed later after your review).

Also, you had removed "inhabited" from an implication which triggered a contradiction in the database. (Good that I have a trigger to check that!) I have added it back in the commit and now the data is consistent again.

Let me know if you are OK, then I will merge.

varkor:
Oops, thank you for fixing those oversights! I must have gotten mixed up about whether I was thinking locally or not 🤦 

I've mentioned the reference I took the terminology from in https://github.com/ScriptRaccoon/CatDat/pull/3#discussion_r2971401898. Looking on Google Scholar, the term does not seem to be frequently used, though it has also been used for a category in which every hom-set is non-empty (which is a slightly stronger condition). However, Bunge's usage of "strongly connected" seems more in keeping with the terminology "connected", which is non-directed.

(Sorry for the late response; the weekend has been quite busy.)

PR #6, opened 2026-03-18
Reference a couple of other relevant projects

varkor:
I would also recommend adding **CatDat** to [MathBases](https://mathbases.org): see their [contribution guide here](https://mathbases.org/contributing).


ScriptRaccoon:
Thank you for adding these!

Regarding Mathbases, I have made a [PR](https://github.com/MathBases/MathBases/pull/134) to add CatDat there.

PR #21, opened 2026-03-26
Add "locally strongly finitely presentable" as a property

ykawase5048:
The property "locally strongly finitely presentable" is the many-sorted extension of the existing property "finitary algebraic".

I also removed "locally ess. small" from the definition of "locally (finitely) presentable" categories, because it follows from the other conditions. (More generally, this still holds for accessible categories.)


ScriptRaccoon:
Ok, I have mentioned this in 8fc6523c7d2c64505986a6f03fce796ad45f5cd9.

PR #39, opened 2026-04-06
Use a separate database for page visits

ScriptRaccoon:
The table with page visits should not be in the same database as the data with all the categories and functors. So this PR creates (and uses) a separate database for these.

Reasons:
- Separation of  concerns
- Preparation for a feature where users can download the database. And of course, they should not download all the visits.

The database backup has also been deactivated for now.

The previous visits will be migrated to the new table.


ScriptRaccoon:
> Preparation for a feature where users can download the database.

This is finally done: #125

PR #54, opened 2026-04-11
Add sifted and cosifted categories

peabrainiac:
Add sifted / cosifted categories and show that filtered categories are sifted, that sifted categories are connected and that categories with finite coproducts are sifted.

These implications were enough to settle siftedness of all but 8 of the categories in the database and cosiftedness of all but 4. I've decided siftedness for most of these remaining examples manually; the only example I didn't decide is the delooping of the additive monoid of ordinal numbers, because I'm a bit confused about the order of composition that is used. The definition of [left_cancellative](https://catdat.app/category-property/left_cancellative) seems to indicate that the composition of $X\overset f\to Y\overset g\to Z$ is denoted $f\circ g$ instead of $g\circ f$; is that a typo, or is that indeed the convention used here?


ScriptRaccoon:
Thank you for the contribution! I will have a look at it. 

I have fixed the typos in the definition of left and right cancellative categories: https://github.com/ScriptRaccoon/CatDat/commit/a58d2bfbb1008e3af743e3a2637ff9e957090490

You can rebase your PR on `main`.

database/data/003_properties/010_misc.sql
```diff
@@ -55,6 +55,22 @@ VALUES
 	'filtered',
 	TRUE
 ),
+(
+	'sifted',
+	'is',
+	'A category $\mathcal C$ is <i>sifted</i> if it is inhabited and the diagonal functor $\Delta:\mathcal C\to\mathcal C\times\mathcal C$ is final, i.e. if it is nonempty and for any two objects $c,d\in\mathcal C$ the category of cospans $c\to e\leftarrow d$ is connected. Equivalently, a small category $\mathcal C$ is sifted if $\textrm{colim}:\textbf{Set}^{\mathcal C}\to\textbf{Set}$ preserves finite products.',
```
ScriptRaccoon: nonempty -> non-empty (see `CONTRIBUTING.md` / Conventions)

database/data/003_properties/010_misc.sql
```diff
@@ -55,6 +55,22 @@ VALUES
...
+	'https://ncatlab.org/nlab/show/sifted+category',
+	'cosifted',
+	TRUE
+),
+(
+	'cosifted',
+	'is',
+	'A category $\mathcal C$ is <i>cosifted</i> if it is inhabited and the diagonal functor $\Delta:\mathcal C\to\mathcal C\times\mathcal C$ is initial, i.e. if it is nonempty and for any two objects $c,d\in\mathcal C$ the category of spans $c\leftarrow e\to d$ is connected. Equivalently, a small category $\mathcal C$ is cosifted if $\textrm{colim}:\textbf{Set}^{{\mathcal C}^\textrm{op}}\to\textbf{Set}$ preserves finite products.',
```
ScriptRaccoon: nonempty -> non-empty (see `CONTRIBUTING.md` / Conventions)

ScriptRaccoon: Shouldn't it say that the diagonal functor is *cofinal*?

ScriptRaccoon:
Remark: I have added a commit that adds related properties (https://github.com/ScriptRaccoon/CatDat/pull/54/changes/115ebd448ea6135fb147765439ab02778ed82610). Maybe I should add this to `CONTRIBUTING.md`.

database/data/004_property-assignments/BG.sql
```diff
@@ -46,4 +46,10 @@ VALUES
 	'essentially finite',
 	FALSE,
 	'This is because we choose $G$ to be infinite.'
+),
+(
+	'BG',
+	'sifted',
+	FALSE,
+	'The category of cospans from $*$ to $*$ in the delooping $BG$ of any group $G$ is equivalent to a discrete category with $|G|$ objects.'
 );
```
ScriptRaccoon: Please remove this and replace it with an implication saying: `groupoid + sifted => trivial`. This can be added to `database/data/005_implications/009_misc-implications.sql`.

database/data/004_property-assignments/BG_f.sql
```diff
@@ -40,4 +40,10 @@ VALUES
 	'trivial',
 	FALSE,
 	'This is trivial.'
+),
+(
+	'BG_f',
+	'sifted',
+	FALSE,
+	'The category of cospans from $*$ to $*$ in the delooping $BG$ of any group $G$ is equivalent to a discrete category with $|G|$ objects.'
 );
```
ScriptRaccoon: This becomes obsolete then (see my other comment on BG).

database/data/004_property-assignments/BN.sql
```diff
@@ -76,4 +76,10 @@ VALUES
 	'one-way',
 	FALSE,
 	'This is trivial.'
+),
+(
+	'BN',
+	'sifted',
+	FALSE,
+	'The category of cospans from $*$ to $*$ in $B\mathbb N$ is equivalent to a countable discrete category.'
```
ScriptRaccoon: I am not sure about that. We have connected components indexed by $\mathbb{Z}$ (and this is already sufficient for the proof), but the components are not ess. discrete, right?

database/data/004_property-assignments/walking_span.sql
ScriptRaccoon: This can be removed now (since filtered => sifted).

database/data/004_property-assignments/FI.sql
```diff
@@ -118,4 +118,10 @@ VALUES
 	'strongly connected',
 	FALSE,
 	'There is no map from a non-empty set to the empty set.'
+),
+(
+	'FI',
+	'sifted',
+	FALSE,
+	'For any two finite sets $X,Y$ and elements $x\in X,y\in Y$ morphisms in the category of cospans $X\overset f\to Z\overset g\leftarrow Y$ preserve whether $f(x)=g(y)$, so e.g. the category of cospans between two singleton sets has exactly two connected components, not one.'
 );
```
ScriptRaccoon: Can we turn this into a general implication? We already have `filtered + left cancellative => thin`. Not sure if we can just replace `filtered` with `sifted` here, but it's worth trying to adapt this result.

cf. https://github.com/ScriptRaccoon/CatDat/pull/54#discussion_r3067837499

EDIT: actually, I think `sifted + left cancellative => thin` *is* true. Consider the connected component of the cospan $X \to X \leftarrow X$... Hence I suggest to replace the current implication `filtered_left_cancellative_implies_thin` with this stronger result (and prove it).

database/data/003_properties/010_misc.sql
peabrainiac: I went with the terminology nlab uses here - [this article](https://ncatlab.org/nlab/show/final+functor) calls functors dual to final functors initial, and explains that this is because final functors used to be called cofinal because of their relation to order theory so that calling the dual notion cofinal now would be ambiguous. A constant functor is also initial iff the corresponding object is, so I think this makes sense.

database/data/004_property-assignments/FI.sql
ScriptRaccoon: Here is the proof:

Let $\mathcal{C}$ be a left cancellative category (every morphism is a monomorphism) which is sifted. I claim that the connected component of the cospan $(\mathrm{id}_X,\mathrm{id}_X)$ from $X$ to $X$ consists of those cospans $(f,f)$ where $f : X \to Y$ is a morphism. First, if $f : X \to Y$ is a morphism, then $f$ is clearly a morphism from the cospan $(\mathrm{id}_X,\mathrm{id}_X)$ to $(f,f)$. Now let $f : X \to Y$ be a morphism. Let $(a : X \to Z,b : X \to Z)$ be a cospan from $X$ to $X$. A morphism of cospans $(f,f) \to (a,b)$ is a morphism $h : Y \to Z$ with $hf=a$ and $hf=b$. Hence, $a=b$. Conversely, a morphism of cospans $(a,b) \to (f,f)$ is a morphism $h : Z \to Y$ with $ha=f$ and $hb=f$. Then $ha=hb$ and hence $a=b$ (using that $\mathcal{C}$ is left cancellative).

peabrainiac:
I've rewritten the code using your suggestions now - that `sifted + left cancellative => thin` implication is a great one, it decided siftedness for all remaining categories but `walking_span`.

Thanks for your review, and more generally for starting this project; I'm excited to see where it will go.

ScriptRaccoon:
> I've rewritten the code using your suggestions now - that `sifted + left cancellative => thin` implication is a great one, it decided siftedness for all remaining categories but `walking_span`.

Great!

I love it when adding a property and a few implication decides it for nearly all categories. This is the best case scenario. It happened a few times already.

> Thanks for your review, and more generally for starting this project; I'm excited to see where it will go.

Thank **you** for the contribution!

I have added a few minor commits and will merge the PR now.

ScriptRaccoon (approved):

PR #81, opened 2026-04-14
Add the simplex category

### LINKED ISSUE #90: Add the simplex category
This is the category whose objects are natural numbers, and whose morphisms n -> m are nondecreasing functions {0, 1, ..., n} -> {0, 1, ..., m}.  (There's also the augmented simplex category where you also allow -1 as a strict initial object.)

---
This issue has been created by **Daniel Schepler** via the submission form on https://catdat.app/categories

ScriptRaccoon:
This PR adds the [simplex category](https://ncatlab.org/nlab/show/simplex+category) and decides all of its properties.

It is equivalent to the category of non-empty finite ordered sets. Since the category of finite ordered sets is already in the database, with all of its properties decided, many proofs could be copied.

The only challenge was to show that it is cosifted (which could *not* be copied and also not be deduced from existing implications). The proof is quite interesting.

proof that Delta is cosifted


dschepler:
It is mono-regular: it's easy to see the monos are the injective maps.  And any subset of {0,...,n} is the equalizer of the two maps {0,...,n}->{0,...,n+1}, one of which is the inclusion map, and the other sends i to the smallest element of S which is >=i, or to n+1 if i is greater than all elements of S.

Using the same construction on the image, you can see the epis are the surjective maps.  So, the category is also balanced.

dschepler:
It seems clear the directed system [0] -> [1] -> [2] -> [3] -> ... of inclusion maps shouldn't have a colimit.  On the other hand, it seems like a codirected system should stabilize at some point and then the value at that point would be the limit.

dschepler:
My guess would be the inclusion map [0] -> [1] doesn't have a pushout with itself since there's nothing to specify which copy of 1 is less than the other one in the pushout.

ScriptRaccoon:
Thanks a lot! I will have a look at this in the next days. Do you feel comfortable adding commits to this PR resp. make your own? @dschepler

dschepler:
> Thanks a lot! I will have a look at this in the next days. Do you feel comfortable adding commits to this PR resp. make your own? @dschepler

I certainly wouldn't feel comfortable making commits without investigating the build and test procedures, and I don't know if I have time for that at the moment.

dschepler:
It doesn't have coreflexive equalizers: the two maps [0] -> [1] have a common left inverse given by the unique map [1] -> [0].

ScriptRaccoon:
I think it is more efficient to just copy the arguments / cite the arguments from **FinOrd** (which already has all properties decided). I am doing this now.

PR #92, opened 2026-04-16
Update foundations

ScriptRaccoon:
Small updates to the text about set-theoretic foundations.
- improve English
- specify that limits and colimits are assumed to be small unless stated otherwise.
- add more examples of collections
- added a visualization below

three-levels-of-size


ykawase5048:
In the definition of well-poweredness, "a set" should be replaced with "a collection with small cardinality".

ScriptRaccoon:
> In the definition of well-poweredness, "a set" should be replaced with "a collection with small cardinality".

Thanks. I have used small families instead. https://github.com/ScriptRaccoon/CatDat/pull/92/changes/c36b9602aff76fefc1ed37335fa692d02aebd6ea

Issue #14, opened 2026-03-20
Add more well-known categories


ScriptRaccoon:
While CatDat already has a couple of well-known categories, some are currently missing. Here are some suggestions. Done categories are moved to the bottom.

Notation | Category | Links | Done 
-- | -- | -- | --
$Grp_{fg}$ | finitely-generated groups |  |  ❌
$Grp_{fp}$ | finitely-presented groups |  |  ❌
$Ab_{c}$ |  countable abelian groups |  |  ❌
$grMod_R$ | Graded modules |  |  ❌
$CGWH$ | compactly generated weak Hausdorff spaces |  |  ❌
$hTop$ | Homotopy category of Top |  |  ❌
$TopGrp$ | Topological groups |  |  ❌
$LCH$ | locally compact hausdorff spaces |  |  ❌
$Ban_c$ | Banach spaces with continuous linear maps | [MSE/1424777](https://math.stackexchange.com/questions/1424777) |  ❌
$TopVect$ | Topological vector spaces |  |  ❌
$NormVect$ | Normed vector spaces |  |  ❌
$CompMet$ | Compact metric spaces | [MSE](https://math.stackexchange.com/questions/91787/) | ❌
$C^*-Alg$ | C*-algebras |  |  ❌
$LawMet$ | Lawvere metric spaces |  |  ❌
$Mod(O_X)$ | Modules over a sheaf of rings |  |  ❌
$RS$ | Ringed spaces |  |  ❌
$Qcoh(X)$ | Quasi-coherent sheaves | [Thesis](https://arxiv.org/abs/1410.1716) |  ❌
$Alg(O_X)$ | Algebras over a sheaf of rings |  |  ❌
$Unif$ | uniform spaces* | [MO](https://mathoverflow.net/questions/95923) |  ❌
$Stoch$ | measurable spaces with Markov kernels | [nlab](https://ncatlab.org/nlab/show/Stoch) |  ❌
$\Pi_1(X)$ | fundamental groupoid of a (specific) space, say $X = S^1$ | |  ❌
$Born$ | bornological sets / spaces || ❌
$Grp_{c}$ | countable groups |  |  ✅
$CompHaus$ | Compact Hausdorff spaces |  |  ✅
$Haus$ | Hausdorff spaces |  |  ✅
$\Delta$ | Simplex category |  |  ✅
$SemiGrp$ | semigroups |  |  ✅
$TorsAb$ | torsion abelian groups |  |  ✅

[...]

PR #95, opened 2026-04-16
Add categories of operator algebras (WIP)

ScriptRaccoon:
It has been suggested in https://github.com/ScriptRaccoon/CatDat/issues/71 to add the category of commutative von Neumann algebras.  Also, https://github.com/ScriptRaccoon/CatDat/issues/14 suggests to add (among other things) the category of C*-algebras. In this PR, various categories of operator algebras will be added.

We will not be able to decide many of their properties (for now).

🤔  Maybe I should make this more focussed and only add Banach algebras in this PR, then have another PR for C*-algebras, then another one for von Neumann algebras. In each case, with all variants: unital / non-unital, commutative / non-commutative.

**TODO.** Shrink this PR down and only add two categories: unital banach algebras, unital commutative banach algebras. The rest can be done in other PRs.

**EDIT.** This PR is paused. First, I want to finish https://github.com/ScriptRaccoon/CatDat/pull/127


ScriptRaccoon:
I will close here, since this PR has been inactive for a very long time and it targets a very old version of CatDat.

The mathematical content has more or less been copied to the linked issues, so that this can be reused.

PR #115, opened 2026-04-19
Congruence quotients

dschepler:
Add property of having quotients of congruence relations, and its dual.

Currently undecided categories for having quotients of congruences:
    none

Currently undecided categories for having coquotients of cocongruences:
    none


dschepler:
For category of metric spaces with continuous maps: I could probably make an argument it would imply having all sequential colimits, and reach a contradiction based on the previously existing database entry.  I would do this by forming the set-level equivalence relation involved in the sequential colimit, showing that this set with the subspace metric forms a congruence, and then showing that congruence's quotient would have to be a sequential colimit.

dschepler:
For the category of schemes, I wonder if I could adapt the example of pushout failure by using the relation $k(x,y) \sqcup Z(x-y)$ in $\mathbb{A}^1 \times \mathbb{A}^1 \simeq Spec ~ k[x,y]$.

databases/catdat/data/004_property-assignments/Met_c.sql
```diff
@@ -100,4 +100,10 @@ VALUES
...
...cting <a href="https://mathoverflow.net/questions/510316" target="_blank">MO/510316</a>. To see this, suppose we have a sequence $X_1 \overset{i_1}\longrightarrow X_2 \overset{i_2}\longrightarrow \cdots$. We then construct $X := \bigsqcup_{n=1}^\infty X_n$ and the set-level relation $R$ on $X$ as in the construction of sequential colimits in $\mathbf{Set}$. Now give $R$ the subspace metric. We see that $R$ is in fact a congruence in $\mathbf{Met}_c$: this is because with this choice of metric, a continuous function $T \to X \times X$ factors through $R$ in $\mathbf{Met}_c$ if and only if its underlying function factors through $R$ in $\mathbf{Set}$, so the induced subset of $\operatorname{Hom}(T, X) \times \operatorname{Hom}(T, X)$ is an equivalence relation for every $T$.  We now claim that a quotient of $R$ must be a sequential colimit of $X_i$. To see this, suppose we have a compatible sequence of maps $f_i : X_i \to Y$.  Then $\sum_i f_i : X \to Y$ is continuous; furthermore, $(\sum_i f_i) \circ \pi_1 \circ \operatorname{inc}_R = (\sum_i f_i) \circ \pi_2 \circ \operatorname{inc}_R$.  We therefore get an induced map $X / R \to Y$; and the uniqueness of this induced map is easy.'
```
ScriptRaccoon: Most of this is formal. Please add an implication which handles this. I guess we need an infinitary distributive category (maybe more).

database/data/005_implications/001_limits-colimits-existence-implications.sql
```diff
@@ -48,6 +48,27 @@ VALUES
...
+	'["quotients of congruences"]',
+	'["finitely complete"]',
+	'This is part of the definition.',
+	FALSE
+),
+(
+	'congruence_quotients_are_reflexive_coequalizers',
+	'["reflexive coequalizers", "finitely complete"]',
+	'["quotients of congruences"]',
+	'For a congruence $i : E \hookrightarrow X \times X$, the two morphisms $p_1 \circ i$ and $p_2 \circ i$ have a common section given by the reflexivity morphism $X \to E$.',
```
ScriptRaccoon: Notice that I have fixed the proof in https://github.com/ScriptRaccoon/CatDat/pull/115/changes/cf835f865e6b19b22eff23cba09233a914a44142

A reflexive pair has a common *section*. This connects to my next comment ...

database/data/005_implications/001_limits-colimits-existence-implications.sql
```diff
@@ -48,6 +48,27 @@ VALUES
...
+	FALSE
+),
+(
+	'reflexive_coequalizers_via_congruence_quotients',
+	'["quotients of congruences", "Malcev"]',
+	'["reflexive coequalizers"]',
+	'Given a parallel pair $f, g : X \to Y$ with a common retraction $r$, then $f$ and $g$ are split monic so together they define a relation $\langle f, g \rangle : X \hookrightarrow Y \times Y$.  This relation is reflexive using $r$, and is therefore a congruence since the category is Malcev.  The quotient of this congruence is then the coequalizer of $f$ and $g$.',
+	FALSE
```
ScriptRaccoon, line 70: I can't follow this proof. A reflexive pair has a common *section*. The morphisms $f$ and $g$ are *epimorphisms*, not *monomorphisms*.

https://ncatlab.org/nlab/show/reflexive+coequalizer

Can we fix this proof? If not, the implication needs to be removed.

ScriptRaccoon:
Indeed the category of schemes does not have all quotients of congruences. The counterexample for missing pushouts from MO/9961 can be adjusted. But I think there is a general lemma behind this. Let me sketch it for now, maybe we can fill in the details.

Lemma: Assume that $C$ is category which is (finitary) distributive and has disjoint finite coproducts. Assume further that it has quotients of congruences. Then any pair of *monomorphisms* $f : S \to X$, $g : S \to Y$ has a pushout.

(Unfortunately, I don't know if all pushouts need to exist.)

Proof: Let $W = X + Y$, so that $W \times W = (X \times X) + (Y \times Y) + (X \times Y) + (Y \times X)$. Let $E := X + Y + S + S$ and define $E \hookrightarrow W \times W$ by $\Delta_X$ on $X$, likewise on $Y$, by $(f,g) :S \to X \times Y$ on the first copy of $S$, and by $(g,f) : S \to Y \times X$ on the second copy of $S$. This is a congruence. Intuitively, it is the smallest congruence satisfying $f(s) \equiv g(s)$. So any quotient is a pushout of $f$ and $g$.

ykawase5048:
Congruences (internal equivalence relations) can be naturally defined in arbitrary categories without finite limits:
A congruence on an object $X$ is a parallel pair $R\rightrightarrows X$ such that the induced morphism $yR \to yX\times yX$ in the presheaf category is monic and is an equivalence relation (pointwise).

ScriptRaccoon:
> Congruences (internal equivalence relations) can be naturally defined in arbitrary categories without finite limits: A congruence on an object X is a parallel pair R ⇉ X such that the induced morphism y R → y X × y X in the presheaf category is monic and is an equivalence relation (pointwise).

I have thought about this as well while reviewing the PR but came to the conclusion that we should keep the assumption here.

dschepler:
With the removal of the incorrect links to reflexive coequalizers, now the case of free abelian groups is undecided.

Of course, if this were the category of torsion-free abelian groups, then we could just take the torsion-free part of the abelian group quotient.  But in this case, maybe there's some theorem about extending a basis of a "saturated" subgroup to a basis of the full group -- "saturated" meaning $nx \in H \rightarrow x \in H$?  In which case the quotient group by the saturation of the subgroup corresponding to the congruence would have basis given by the images of the extension vectors, and it would form a quotient in FreeAb.

On the other hand, we again have an existing database entry about not having sequential colimits...

dschepler:
Maybe a common lemma, to be used for both Met_c and FreeAb, could be something like: Suppose you have a faithful (and exact?) functor U : C -> Set, and suppose furthermore that for each object X of C and subset S of U(X), there exists a subobject S' of X such that (fill in what's needed here).  Then if C has quotients of congruences, then C has sequential colimits.  (In the case of Met_c, that subobject would be the subspace metric construction; in the case of FreeAb, that subobject would be the generated subgroup, which is free again by the theorem about subgroups of free groups.)

dschepler:
On the other hand, maybe I could put in the simple implication preadditive + has quotients of congruences -> has cokernels which would also resolve FreeAb.

ykawase5048:
> > Congruences (internal equivalence relations) can be naturally defined in arbitrary categories without finite limits: A congruence on an object X is a parallel pair R ⇉ X such that the induced morphism y R → y X × y X in the presheaf category is monic and is an equivalence relation (pointwise).
> 
> I have thought about this as well while reviewing the PR but came to the conclusion that we should keep the assumption here.

In multi-algebraic categories, every congruence admits a coequalizer, in which finite limits don't exist in general. We may find other examples where congruences without finite limits play an important role. I strongly recommend adopting the general definition.

dschepler:
> > > Congruences (internal equivalence relations) can be naturally defined in arbitrary categories without finite limits: A congruence on an object X is a parallel pair R ⇉ X such that the induced morphism y R → y X × y X in the presheaf category is monic and is an equivalence relation (pointwise).
> > 
> > 
> > I have thought about this as well while reviewing the PR but came to the conclusion that we should keep the assumption here.
> 
> In multi-algebraic categories, every congruence admits a coequalizer, in which finite limits don't exist in general. We may find other examples where congruences without finite limits play an important role. I strongly recommend adopting the general definition.

Maybe we could make that into another property for those situations - something like "has representable quotients of Yoneda embeddings".  (And off the top of my head, I wouldn't expect that to be equivalent to the current property even in finitely complete categories, since the corresponding subfunctors of Hom(-,X) x Hom(-,X) wouldn't necessarily be representable.)

ykawase5048:
> (And off the top of my head, I wouldn't expect that to be equivalent to the current property even in finitely complete categories, since the corresponding subfunctors of Hom(-,X) x Hom(-,X) wouldn't necessarily be representable.)

We define a (general) congruence as a special parallel pair of morphisms, so the corresponding subfunctor is trivially representable.

dschepler:
> We define a (general) congruence as a special parallel pair of morphisms, so the corresponding subfunctor is trivially representable.

Oh, right.  So, that could be "has quotients of congruence pairs".  Or, if @ScriptRaccoon changes their mind, I'd be happy to use the more general definition (and see what new unresolved categories get exposed - in that case, I could need some help with the implications you've mentioned related to multi-algebraic categories).

dschepler:
> Indeed the category of schemes does not have all quotients of congruences. The counterexample for missing pushouts from MO/9961 can be adjusted. But I think there is a general lemma behind this. Let me sketch it for now, maybe we can fill in the details.
> 
> Lemma: Assume that C is category which is (finitary) distributive and has disjoint finite coproducts. Assume further that it has quotients of congruences. Then any pair of _monomorphisms_ f : S → X , g : S → Y has a pushout.
> 
> (Unfortunately, I don't know if all pushouts need to exist.)
> 
> Proof: Let W = X + Y , so that W × W = ( X × X ) + ( Y × Y ) + ( X × Y ) + ( Y × X ) . Let E := X + Y + S + S and define E ↪ W × W by Δ X on X , likewise on Y , by ( f , g ) : S → X × Y on the first copy of S , and by ( g , f ) : S → Y × X on the second copy of S . This is a congruence. Intuitively, it is the smallest congruence satisfying f ( s ) ≡ g ( s ) . So any quotient is a pushout of f and g .

I think we might need to add extensivity to prove transitivity, since for that one pullbacks are involved.

So yes, eventually I think we could put an implication "extensive + quotients of congruences -> pushouts of monomorphisms" - though I certainly wouldn't want to introduce a new property to this PR and risk it growing out of scope like the previous one.

ScriptRaccoon:
> In multi-algebraic categories, every congruence admits a coequalizer, in which finite limits don't exist in general. We may find other examples where congruences without finite limits play an important role. I strongly recommend adopting the general definition.

My idea was that assuming finite limits is in alignment with many other properties where other properties are assumed beforehand even though they are not strictly necessary - it just makes things more convenient. Typical example: "exact filtered colimits". Here I assume that both finite limits and filtered colimits exist. (For a good reason, otherwise the 3 equivalent characterizations would not be equivalent anymore!)

But I trust your expertise here when you say it is better to omit finite limits here. And you can also change the definition now @dschepler  and see how the PR evolves. This is the nice thing about git, we can always go back (unless one force pushes all the time like myself ...).

> So, that could be "has quotients of congruence pairs". 

For now, let's not have two properties in the database which sound very similar.
 
 > So yes, eventually I think we could put an implication "extensive + quotients of congruences -> pushouts of monomorphisms" - though I certainly wouldn't want to introduce a new property to this PR and risk it growing out of scope like the previous one.
 
 For this, we have lemmas, see https://github.com/ScriptRaccoon/CatDat/pull/41. Here is a [recent example](https://catdat.app/lemma/filtered-monos).

dschepler:
I just tried a quick experiment removing the implication "quotients of congruences -> finitely complete" and removing the hypothesis of finitely complete where appropriate (for example, I couldn't remove it from the implication of congruence quotients -> cokernels because it depends on constructing a kernel of a map from $X^2$).  Once I added back in the implication "reflexive coequalizers -> quotients of congruences" which I guess I was too hasty to remove, that leaves as undecided cases:
For quotients of congruences: category of sets and relations, category of smooth manifolds.
For coquotients of cocongruences: category of non-empty sets, category of sets and relations, category of smooth manifolds, simplex category.

(For smooth manifolds, it's of course tempting to try the congruence identifying two points of a line.  But as usual, just because the topological space quotient isn't a manifold doesn't mean there isn't a quotient in the subcategory.  For example, it could be that $S^1$ works as a quotient.)

databases/catdat/data/003_properties/002_limits-colimits-existence.sql
```diff
@@ -327,6 +327,22 @@ VALUES
 	'coreflexive equalizers',
 	TRUE
 ),
+(
+	'quotients of congruences',
+	'has',
+	'A <i>congruence</i> (or internal equivalence relation) on an object $X$ of a category is a parallel pair $p_1, p_2 : E \rightrightarrows X$ which is jointly monomorphic, and such that for every object $T$, the image of $(p_1 \circ {-}, p_2 \circ {-}) : \operatorname{Hom}(T, E) \to \operatorname{Hom}(T, X)^2$ is an equivalence relation. The category has <i>quotients of congruences</i> if for each such congruence, there exists a coequalizer of $p_1$ and $p_2$. Note that in the case of a category with binary powers, the corresponding subobjects of $X \times X$ are also commonly referred to as congruences, or as internal equivalence relations.',
```
ScriptRaccoon, line 333: There is a typo here*. I recommend to install the recommended extensions in `extensions.json` if you are using vscode. This includes the spell checker which uses an english dictionary + the special words defined in `settings.json`.

Image<br>
 
*sorry I mean in the def of coquotients of cocongruences below.

databases/catdat/data/010_lemmas/000_lemmas.sql
```diff
@@ -114,4 +114,10 @@ INSERT INTO lemmas (
     For the record, here is the dual statement: Let $\mathcal{C}$ be a category with cofiltered limits. Assume that $U : \mathcal{C} \to \mathcal{D}$ is faithful functor which preserves epimorphisms and cofiltered limits. If epimorphisms in $\mathcal{D}$ are stable under cofiltered limits, then the same is true for $\mathcal{C}$.
     ',
     'Since $U$ is faithful, it reflects monomorphisms. From here the proof is straight forward.'
-);
\ No newline at end of file
+),
+(
+    'pushouts-of-monos-via-congruence-quotients',
```
ScriptRaccoon, line 119: Cool! Thanks for adding this. I was hoping that you could fill in the details for my proof sketch. Your proof is even shorter than my sketch :D

ScriptRaccoon:
Looks very good already! Maybe we can settle the remaining cases as well. 

If time permits, please have a look at the open comments.

dschepler:
I just realized that the "Malcev" property also involves congruences and in that case the decision was to assume the category was finitely complete.

ScriptRaccoon:
> I just realized that the "Malcev" property also involves congruences and in that case the decision was to assume the category was finitely complete.

Yes. This is part of Definition 2.2.3 in Borceux-Bourn.

dschepler:
Well, I ended up showing RelSet has quotients of congruences, but the proof was fairly long.  It's the sort of proof where you gather bits of information little by little, until eventually you can put together enough to classify the congruences (and there aren't that many of them).  Given how long it turned out, I'm wondering if it might make sense to move it to "lemmas" even though it's probably not of use for anything else.

dschepler:
At this point, the only unresolved case is: does Man have coquotients of cocongruences.  Given the picture of a cocongruence on X in a concrete category looking like a branching of X into one or two parts at various points, I'd strongly suspect that means the number of branches is locally constant, so the equalizer where you have one part is an open subset of X.  From the Hausdorff property of E it's easy to see where you have two parts is open but that doesn't help.

dschepler:
That would also leave the proof for Met_c which I've been thinking about but haven't yet found a good way to generalize the argument.

ScriptRaccoon:
@dschepler  @ykawase5048  Remark: after I merge https://github.com/ScriptRaccoon/CatDat/pull/119 and you pull the change, you need to update your local `.env` file (see the updated `.env.example` as a guideline), since location of the local databases will be different. I will inform you when I merge.

dschepler:
> In multi-algebraic categories, every congruence admits a coequalizer...

Does that mean it would be appropriate to add an implication "multi-algebraic -> has quotients of congruences"?  And if so, is there a citation you can provide for the proof entry?  (Though it doesn't look like it would add any resolved entries in the current state of the database, it could be useful in the background for future categories.)

dschepler:
I've now filled in an entry for Man on coquotients of cocongruences.  With that, all current categories are settled for the new properties.

dschepler:
> > In multi-algebraic categories, every congruence admits a coequalizer...
> 
> Does that mean it would be appropriate to add an implication "multi-algebraic -> has quotients of congruences"? And if so, is there a citation you can provide for the proof entry? (Though it doesn't look like it would add any resolved entries in the current state of the database, it could be useful in the background for future categories.)

Actually, I guess we already have the chain of implications multi-algebraic -> generalized variety -> sifted colimits -> reflexive coequalizers -> quotients of congruences so we don't need to add an explicit implication there.

dschepler:
Now I'm worried... The proof I found for Man would seem to apply for any coreflexive equalizer, but the implications say Man doesn't have general coreflexive equalizers.  So, I may have to trace through the implications from a failure to have pullbacks, in order to see whether my proof is flawed, or if I'm OK and I just missed something I used that's not true in the general coreflexive equalizer case.  It might be a while before I can get enough time to devote to the detailed analysis, though.

dschepler:
OK, in the general case of a coreflexive equalizer $X \rightrightarrows Y$ you don't necessarily require $X + X \to Y$ to be an epimorphism, so in the case of Man, Y could certainly have higher dimension than X.  On the other hand, I forgot that the epimorphism only has to have a dense image at the level of sets, so I'm not sure whether the part about fibers of one or two points would necessarily hold.  Given the setup it seems unlikely to be able to have something like a "spirograph curve" as the epimorphism, but I'd have to think about that some more.

ScriptRaccoon:
Since #119 is now merged, a big rebase is required. (I can also do it if you prefer.) **EDIT.** I just did this locally and it went through without any merge conflict. 🙏🏻 So I think you can do it easily.

Also, because of the new location of the local database files, `.env` needs to be adjusted (this is not part of the repo, hence this message).

database/data/004_property-assignments/Delta.sql
```diff
@@ -75,6 +75,12 @@ VALUES
     <p>$\begin{array}{ccccc} X & \xleftarrow{x_0} & [0] & \xrightarrow{y} & Y \\ \parallel && \phantom{ {\footnotesize 0}} \downarrow {\footnotesize 0} && \parallel \\ X & \xleftarrow{f} & [1] & \xrightarrow{g} & Y \\ \parallel && \phantom{ {\footnotesize 1}} \uparrow  {\footnotesize 1} && \parallel \\ X & \xleftarrow{x_1} & [0] & \xrightarrow{y} & Y \end{array}$</p>
     This shows that the choice of $x \in X$ does not matter, and for $y \in Y$ the proof is the same.'
 ),
+(
+    'Delta',
+    'coquotients of cocongruences',
+    TRUE,
+    'Given a jointly epimorphic pair of maps $f, g : [m] \to [n]$, that means the union of the images is all of $[n]$.  Also, given the coreflexivity map $r : [n] \to [m]$ with $r \circ f = r \circ g = \operatorname{id}$, we must have that $f(i)$ and $g(i)$ map to adjacent or equal elements for each $i \in [m]$.  Taking into consideration that the cosymmetry map $s : [n] \to [n]$ would have to swap the elements in the adjacent case, giving a contradiction to $s$ being nondecreasing, we see that in fact $f = g$.  Therefore, $[m]$ is the equalizer of $f$ and $g$.'
```
ScriptRaccoon: I must say that I don't understand this proof. 
e.g. why must f(i), g(i) be adjacent?

ScriptRaccoon:
> Well, I ended up showing RelSet has quotients of congruences, but the proof was fairly long. It's the sort of proof where you gather bits of information little by little, until eventually you can put together enough to classify the congruences (and there aren't that many of them). Given how long it turned out, I'm wondering if it might make sense to move it to "lemmas" even though it's probably not of use for anything else.

I will have a look at this proof. For now, let me just remark that there is also an option to just put a pdf into the `/static/pdf` folder and link to it. This can be used for longer proofs, or proofs which require complicated diagrams. This has been done once before: https://github.com/ScriptRaccoon/CatDat/pull/43

(I am not happy that we have basically four ways to add proofs: property assignment reasons, implication reasons, lemmas, pdfs. And for the foundations I also built a markdown renderer. There should be just one way ... But for now it is what it is.)

dschepler:
For now I've backed out the Man assignment, pending a fix to the proof (or a completely new proof if needed).  That means that again, the status of Man and coquotients of cocongruences is unresolved.

dschepler:
> For now I've backed out the Man assignment, pending a fix to the proof (or a completely new proof if needed). That means that again, the status of Man and coquotients of cocongruences is unresolved.

That's fixed again now with a preliminary paragraph plugging the gap in the proof.

ScriptRaccoon:
@dschepler Sounds great! Can you please let me know as soon as you think the PR is finished? Then I will review it again. For now, I will put it back to draft mode since you are still working it (as far as I can tell).

dschepler:
> @dschepler Sounds great! Can you please let me know as soon as you think the PR is finished? Then I will review it again. For now, I will put it back to draft mode since you are still working it (as far as I can tell).

I think I'm done with it now, pending further review comments of course.

ScriptRaccoon:
Alright.

I have looked again over all the changes. I made a commit in which I have simplified the proofs for $\Delta$ and $Man$, and also adjusted the formatting a bit. Please have a look: https://github.com/ScriptRaccoon/CatDat/pull/115/changes/0093b62a2e3360eaf66d1d09c1759acce95eaf56

Regarding $\Delta$, we might want to add the property https://ncatlab.org/nlab/show/gaunt+category at some point. Every gaunt category has quotients of congruences for trivial reasons. EDIT: https://github.com/ScriptRaccoon/CatDat/pull/121

I still think that it would be nice to have more details in the proofs of the two lemmas, but these details can also be added later. I think that extensivity is what enables us to copy the usual proof for the category of sets.

I am quite surprised that *exactly one* category in the database does not have coquotients of cocongruences, the category non-empty sets.

Oups, as of writing this, I have realized that I still need to look at the proof for **Rel**. Notice that we need to change this since the proof is not readable on mobile devices, it overflows. Can you create a pdf and add it to `/static/pdf`?

*EDIT.* Or can we find a paper which classifies the congruences as you did?

*EDIT2.* I have asked on MO: https://mathoverflow.net/questions/510539

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,27 @@
...
+\usepackage{amsmath, amssymb, amsfonts, mathtools}
+
+\title{The category of sets and relations has quotients of congruences}
+\author{Daniel Schepler}
+\date{\today}
+
+\begin{document}
+\maketitle
+
+Let $i : E \hookrightarrow X \times X$ be a congruence.  Then:
```
ScriptRaccoon: Maybe it is easier to classify cocongruences? Because that product is really a disjoint union of sets. Or how do you "visualize" $i$?

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,27 @@
...
+\begin{itemize}
+\item The corresponding function $i_* : P(E) \to P(X) \times P(X)$ must be injective, preserve arbitrary unions and in particular be inclusion-preserving, and have image equal to an equivalence relation $\sim$ on $P(X)$.  Note also that the fact that $i_*$ preserves arbitrary unions implies that $\sim$ respects arbitrary unions.
+\item Since the symmetry morphism $s : E \to E$ satisfies $s^2 = \operatorname{id}$, it must be the graph of an involution $s_0$ on $E$.
```
ScriptRaccoon: If you are using that isomorphisms are functions (not sure atm), you can add that this uses the classification of isomorphisms (below on the page).

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,27 @@
...
+\item For the reflexivity morphism $r : X \to E$, we must have $r_*$ of each singleton $\{ x \}$ must be either a singleton or a pair.  To see this, $r_*(\{ x \})$ certainly cannot be empty, or its image under $i_*$ would be $(\emptyset, \emptyset)$, contradicting that $i_* \circ r_*$ is the diagonal map.  If $r_*(\{ x \})$ had more than two elements, then choose a chain of subsets with four elements.  That must map under $i_*$ to a chain of subsets of $(\{ x \}, \{ x \})$ with each image distinct, giving a contradiction.
```
ScriptRaccoon: I don't understand this sentence 

> If $r_*(\{ x \})$ had more than two elements, then choose a chain of subsets with four elements

and neither what follows. Do you mean four subsets maybe? Personally, I would be able to follow this proof a bit more easily if there were more formulas.

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,27 @@
...
+\item Similarly, if $r_*(\{ x \})$ is a pair, then one element of the pair must map under $i_*$ to $(\emptyset, \{ x \})$ and the other must map to $(\{ x \}, \emptyset)$.  Since $s \circ r = r$ and $i \circ s = (p_2, p_1) \circ i$, we must have that $s_0$ maps each element to the other.  Otherwise, if $r_*(\{ x \})$ is a singleton, then the element of that singleton is a fixed point of $s_0$.
```
ScriptRaccoon: Again, I cannot follow.

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,27 @@
...
+\item Suppose that under the equivalence relation on $P(X)$, we have $S \sim \emptyset$.  Then for any $x \in S$, we have $\{ x \} = \{ x \} \cup \emptyset \sim \{ x \} \cup S = S \sim \emptyset$.  Using the fact that $i_*$ preserves unions for the inverse direction, this shows that $S \sim \emptyset$ if and only if $\{ x \} \sim \emptyset$ for every $x \in S$.
+\item Suppose that $r_*(\{ x \}) = \{ e \}$.  We then claim that $\{ x \} \not\sim \emptyset$.  If so, then there would have to be some subset of $E$ which maps to $(\emptyset, \{ x \})$, and similarly to previous steps, this implies that this subset of $E$ would have to be a singleton $\{ f_1 \}$.  Similarly, $(\{ x \}, \emptyset)$ would have to be the image of a singleton $\{ f_2 \}$.  But then $i_*(\{ f_1, f_2 \}) = (\{ x \}, \{ x \}) = i_*(\{ e \})$, contradicting the injectivity of $i_*$.
```
ScriptRaccoon: > and similarly to previous steps

Can you expand this?

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,27 @@
...
+\item Now suppose that $i_*(\{ e \}) = (S, T)$ for $e \in E$.  Then $i_*(\{ s_0(e) \}) = (T, S)$, so $i_*(\{ e, s_0(e) \}) = (S \cup T, S\cup T) = i_*(\bigcup_{x\in S\cup T} r_*(\{ x \})$.  It follows that $\bigcup_{x\in S\cup T} r_*(\{ x \}) = \{ e, s_0(e) \}$.  Therefore, $E$ is covered by all the sets $r_*(\{ x \})$.
```
ScriptRaccoon: nit: Can you use `\[ ... \]` here, in particular to remove the overflow?

ScriptRaccoon: What is $s_0$? Maybe you mean $s$?

ScriptRaccoon: After the `r_*(\{ x \})` a bracket is missing.

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,27 @@
...
+\item In summary, we can conclude that the equivalence relation on $P(X)$ is that $S\sim T$ if and only if $S\cap A = T\cap A$ where $A$ is the set of $x\in X$ such that $r_*(\{ x \})$ is a singleton.  This implies that $A$ is a quotient for the congruence, with quotient morphism $X \to A$ given by the relation $\Delta_A \subseteq A \times A \subseteq X\times A$.  (This last point might be easiest to see using the equivalence of $\mathbf{Rel}$ with the Kleisli category of the covariant powerset monad on $\mathbf{Set}$.  This can also be viewed as the category of {\em free} complete $\vee$-semilattices, i.e.~those complete $\vee$-semilattices of the form $P(X)$ for some set $X$.)
```
ScriptRaccoon: > In summary, we can conclude that

I don't see how the previous bullet points lead to that conclusion.

static/pdf/congruences_in_rel.tex
ScriptRaccoon: I think it is better to begin the proof with "We will classify the congruences on a set in $\mathbf{Rel}$."

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,27 @@
...
+
+\title{The category of sets and relations has quotients of congruences}
+\author{Daniel Schepler}
+\date{\today}
+
+\begin{document}
+\maketitle
+
+Let $i : E \hookrightarrow X \times X$ be a congruence.  Then:
+\begin{itemize}
```
ScriptRaccoon: I suggest to replace the bullet points with paragraphs.

ScriptRaccoon (changes requested):
I have a hard time understanding the proof for **Rel**. Here I have added some questions. Maybe you can clarify the proof and make it easier to understand.

Remark: the comments contexts are best seen in the Files tab.

static/pdf/congruences_in_rel.tex
dschepler: After the revision I don't see overflow anymore.  If you have suggestions for where to put \[ ... \] in the revised version, I would need the requests to be more specific.

dschepler:
I wonder... Might it be the case that a congruence, with a quotient which is a split coequalizer, is automatically the kernel pair of the quotient map?

Edit: I guess so. The idea of a split coequalizer is that it defines a retraction $\rho$ of X to an isomorphic copy of the quotient, and a map which gives $x \sim \rho(x)$ for each $x$.  So if $\rho(x_1) = \rho(x_2)$ and you have a congruence, then $x_1\sim \rho(x_1)\sim \rho(x_2)\sim x_2$ so $(x_1,x_2)\in E$.  Now just make it about generalized elements, and pairs of maps factoring through $E$, and that should be a proof.

static/pdf/congruences_in_rel.tex
ScriptRaccoon, line 1: Please replace `\emptyset` by `\varnothing` (see `CONTRIBUTING.md`).

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,55 @@
...
+\author{Daniel Schepler}
+\date{\today}
+
+\begin{document}
+\maketitle
+
+We will give a classification of congruences $i : E \hookrightarrow X \times X$ in the category $\mathbf{Rel}$ of sets and relations.  Recall that products in $\mathbf{Rel}$ are actually given by disjoint unions at the level of underlying sets.  Also, recall that a morphism $R : X \to Y$ in $\mathbf{Rel}$ is a monomorphism if and only if $R_* : P(X) \to P(Y), S \mapsto \{ y\in Y \mid \exists x\in S, (x, y) \in R \}$ is injective, and it is an isomorphism if and only if $R$ is the graph of a bijection $X \to Y$ in $\mathbf{Set}$.
+
+In particular, we get $i_* : P(E) \to P(X + X) \simeq P(X) \times P(X)$ which must be injective.  It must also preserve arbitrary unions and in particular be inclusion-preserving.  From the assumption that $i$ is a congruence, since the functor $(P, {-}_*) : \mathbf{Rel} \to \mathbf{Set}$ functor is representable by the object 1, we see that $i_*$ must have image given by an equivalence relation $\sim$ on $P(X)$.  Note also that since $i_*$ preserves arbitrary unions, we must have that $\sim$ respects arbitrary unions.
```
ScriptRaccoon: typo:

```suggestion
In particular, we get $i_* : P(E) \to P(X + X) \simeq P(X) \times P(X)$ which must be injective.  It must also preserve arbitrary unions and in particular be inclusion-preserving.  From the assumption that $i$ is a congruence, since the functor $(P, {-}_*) : \mathbf{Rel} \to \mathbf{Set}$ is representable by the object 1, we see that $i_*$ must have image given by an equivalence relation $\sim$ on $P(X)$.  Note also that since $i_*$ preserves arbitrary unions, we must have that $\sim$ respects arbitrary unions.
```

static/pdf/congruences_in_rel.tex
```diff
@@ -0,0 +1,55 @@
...
+
+Now suppose that $i_*(\{ e \}) = (S, T)$ for $e \in E$.  Then $i_*(\{ s_0(e) \}) = (T, S)$, so $i_*(\{ e, s_0(e) \}) = (S \cup T, S\cup T) = i_*(\bigcup_{x\in S\cup T} r_*(\{ x \}))$.  It follows that $\bigcup_{x\in S\cup T} r_*(\{ x \}) = \{ e, s_0(e) \}$.  Therefore, $E$ is covered by all the sets $r_*(\{ x \})$.
+
+We now claim that for $S, T \in P(X)$, $S \sim T$ if and only if $S\cap A = T\cap A$, where $A$ is the set of $x\in X$ such that $r_*(\{ x \})$ is a singleton.  For the forward direction, suppose $(S, T) = i_*(U)$ for $U \subseteq E$.  Then each element $e\in U$ is in $r_*(\{ x \})$ for some $x \in X$, so $i_*(\{ e \})$ is equal to one of $(\emptyset, \{ x \}), (\{ x \}, \emptyset), (\{ x \}, \{ x \})$, with $x\notin A$ if either of the first two values holds.  Therefore, $i_*(U) = \bigcup_{e\in U} i_*(\{ e \})$ satisfies the condition.  For the reverse implication, whenever $x\notin A$, we have $\{ x \} \sim \emptyset$.  Therefore, since $S = (S \cap A) \cup \bigcup_{x\in S \setminus A} \{ x \}$, we must have $S \sim S \cap A$.  Similarly, $T \sim T \cap A$, so if $S\cap A = T\cap A$, then $S \sim T$.
```
ScriptRaccoon: Thank you for the revision, but I still do not understand what is happening here. Starting from "... with $x \notin A$ if either ..." (why?). The rest of the paragraph is also unclear to me.

static/pdf/congruences_in_rel.tex
ScriptRaccoon, line 1: Can you perhaps restructure the proof into several claims and make clear when you use which claim? I think this will make it easier to follow.

```
\textbf{Claim 1.} For each $x\in X$, the set $r_*(\{ x \})$ must be either a singleton or a pair.

To see this, ...
```

ScriptRaccoon (approved):

ScriptRaccoon:
Thank you for all the work in this PR; and in particular for all the clarifications you added based on my comments (sorry that there were so many, apparently these days I am not able to understand elementary set theory).

I think we can merge now, right? Or is there anything you would like to add?

The [mathoverflow request](https://mathoverflow.net/questions/510539) hasn't been answered, yet. But if something comes in, we can make a follow-up PR.

dschepler:
> Thank you for all the work in this PR; and in particular for all the clarifications you added based on my comments (sorry that there were so many, apparently these days I am not able to understand elementary set theory).

No problem, I very much like this advice I've seen: if a reviewer doesn't understand something, it's most likely that you need to make edits to clarify, or else future readers might have the same issues.  Your comments helped drastically improve the presentation.

> I think we can merge now, right?

I think so, yes.

ScriptRaccoon:
There is an answer on MO now. https://mathoverflow.net/a/510711/2841

I don't understand it (I usually don't understand proofs which leave out all the details), but maybe you can make something out of it? @dschepler

dschepler:
It looks like essentially the same proof as I gave, just without many details as you said.

PR #126, opened 2026-04-23
Has effective (co)congruences properties

### LINKED ISSUE #114: Does regular + epi-regular + distributive imply co-Malcev?
I was thinking about the theorem that an elementary topos is co-Malcev, and how I would prove it -- and the proof I came up with seems to only require the hypotheses of being regular, epi-regular, and distributive.  The idea is: given a reflexive corelation $X \sqcup X \to R \to X$, use the hypotheses to express the first map as the coequalizer of its kernel pair.  By distributivity, that kernel pair is equivalent to a subobject of $(X \times X)^{\sqcup 4}$, which can be described semi-pictorially as: in the lower left and upper right quadrants, $\Delta_X$, and in the lower right and upper left quadrants, $\Delta_E$ where $E$ is the equalizer of the two maps $X \to R$.  Using that, and the expression of $R$ as a coequalizer of $X \sqcup X$, it should then be straightforward to construct the required symmetry morphism $R \to R$ and the required transitivity morphism from $R$ to the pushout.

And a very rough sketch of the picture I have in my mind of the situation:
```
   -----------
   -----------   <- X u X
         |
         V
   ---\   /---         <- X_1                      <- R
       ----  <- E      <- X_1      <- X_2          <- R
   ---/   \---                     <- X_2          <- R
         |
         V
   -----------     <- X
```
(And the pushout would be the middle diagram but with each branching part repeated three times instead of twice.  Then the symmetry morphism looks like "flipping R upside down" and the transitivity morphism looks like "sending R to the first and third branches".)

Anyway, I also did a quick search through the database and found no counterexamples to dissuade me of the validity of the argument.

dschepler:
I just started a trial run of the property of having effective congruences; and so far, it's not going well.  I only found a couple basic properties to put in, along with a preliminary version of the theorem that a pretopos is balanced; but there are still 34 unresolved cases for congruences and 50 unresolved cases for cocongruences.  I don't even know whether Group has effective cocongruences.  And certainly, there are a lot of cases I could fill in by hand, but that would be a lot of individual entries to maintain.

Any ideas would be welcome on how to proceed.

(I know this is still draft and has several places that need details or citations filled in; at this point I'm just posting to give an idea of the current status.)


databases/catdat/data/003_properties/002_limits-colimits-existence.sql
```diff
@@ -343,6 +343,22 @@ VALUES
 	'quotients of congruences',
 	TRUE
 ),
+(
+	'effective congruences',
+	'has',
+	'A congruence $f, g : E \rightrightarrows X$ (see definition <a href="/category-property/quotients_of_congruences">here</a>) is <i>effective</i> if it is the kernel pair of some morphism, i.e. if there is a morphism $h : X \to Y$ such that we have a cartesian square [put diagram here]. A category <i>has effective congruences</i> if every congruence in the category is effective.',
```
ScriptRaccoon: I read that katex should support amscd, but it doesn't work and there is no proper documentation.

For now, the following works.

```
<p>$\begin{array}{ccc}E & \xrightarrow{f} & X \\ {\footnotesize g} \downarrow \phantom{-} && \phantom{-} \downarrow {\footnotesize h} \\ X & \xrightarrow{h} & Y \end{array}$</p>
```

ScriptRaccoon: ah I just found a way to make it work, but it will require some changes to the rendering functions ...

ScriptRaccoon: https://github.com/ScriptRaccoon/CatDat/pull/128 🚀

After rebasing you can use diagrams now.

dschepler:
Current status: For "has effective congruences" there are two unresolved cases left:
    [category of Banach spaces with linear contractions](http://localhost:5173/category/Ban)
    [category of pseudo-metric spaces with non-expansive maps](http://localhost:5173/category/PMet)

(For the second, I think I might be able to adapt the proof of extensive + has effective congruences -> balanced, by considering $(0, 1) \hookrightarrow [0, 1]$ to bound distances within components, then placing copies "far enough" from each other for the construction to still work even if it's not making a coproduct.  I'd still have to think about the details and whether they work out.  For the first, obviously that trick won't work.)

For "has effective cocongruences" there are still 21 unresolved cases.  Among them are Group and Ring which are blockers.

databases/catdat/data/005_implications/001_limits-colimits-existence-implications.sql
```diff
@@ -69,6 +69,27 @@ VALUES
...
+	'["thin"]',
+	'["effective congruences"]',
+	'In a thin category, any congruence pair is a reflexive fork, and thus consists of equal isomorphisms. Therefore, the congruence is the kernel pair of the identity morphism on the target.',
+	FALSE
+),
+(
+	'left_cancellative_effective_congruences',
+	'["left cancellative"]',
+	'["effective congruences"]',
+	'For any congruence $f, g : E \rightrightarrows X$, if $r$ is the reflexivity morphism we have $r$ is both an epimorphism and a split monomorphism, so it is an isomorphism. Therefore, $f\circ r = g\circ r = \mathrm{id}_X$ implies $f = g$.  As <a href="core-hin_quotients">here</a>, that implies that $E$ is the kernel pair of $\mathrm{id}_X$.',
```
ScriptRaccoon: We have $f r = \mathrm{id}_X$, so $r$ is a split monomorphism. Why is $r$ an epimorphism?

I think what you want to say is that $f$ is a split epimorphism and a monomorphism by assumption, hence an isomorphism. Hence, $f = r^{-1}$. Likewise, $g = r^{-1}$.

ScriptRaccoon:
I am not surprised that deciding effective cocongruences for concrete categories is so hard. This amounts to a classification of all cocongruences, and this is hard, as we already saw in **Rel** for example, but also **Set** is a good starting point, where it is not trivial. Often we do not even understand all epimorphisms.

I suggest that in this PR we only try to fill the remaining cases where it is required by the unit tests (**Grp** and **Ring**).

*EDIT.* I am pretty confident that for **Grp** the answer is yes, cocongruences are effective.

databases/catdat/data/005_implications/001_limits-colimits-existence-implications.sql
dschepler: In the proof that left cancellative implies both reflexive coequalizers and coreflexive equalizers, it already says that in both cases the maps are equal isomorphisms.  Maybe I should just cite that, or even tag the extra conclusions onto that implication.

ScriptRaccoon: The 2nd sounds good.

dschepler:
Heh, ended up coming back to #114 and also using it to prove elementary topoi have effective cocongruences.

ykawase5048:
**_Suggestion._** The implication "lsfp → effective congruences" can be refined by "multi-algebraic → effective congruences". Note that the database already includes the implication "lsfp → multi-algebraic". The reference is Thm. 4.0 in [Diers's paper (fr)](https://doi.org/10.1007/BF01224953) or [it's English translation](https://translations.thosgood.net/ADM-34-1980-193.pdf).

dschepler:
> I am pretty confident that for **Grp** the answer is yes, cocongruences are effective.

Do you have any ideas on how we might prove that?  So far, I haven't made much progress even on the simplest case I can think of, proving that a cocongruence on $\mathbb{Z}$, $F_2 \twoheadrightarrow E$, necessarily has kernel normally generated by $a^n b^{-n}$ for some $n$.

dschepler:
On **Ring** I was wondering if the counterexample in **CommRing** could be adapted there. If I trace through the proofs, I guess the counterexample in **CommRing** is something like $\mathbb{Z} \times \mathbb{Q} \times \mathbb{Q} \times \mathbb{Z}$ as a cocongruence on $\mathbb{Z} \times \mathbb{Z}$. But then again, if you have two maps $\mathbb{Z} \times \mathbb{Q} \times \mathbb{Q} \times \mathbb{Z} \to R$ then the corresponding idempotents of $R$ don't necessarily have an idempotent as product, so that might not extend to a cocongruence on **Ring**.

Any other ideas on **Ring**?

ScriptRaccoon:
> Do you have any ideas on how we might prove that? So far, I haven't made much progress even on the simplest case I can think of, proving that a cocongruence on Z , F 2 ↠ E , necessarily has kernel normally generated by a n b − n for some n .

I don't have a proof for **Grp**, I just voiced my strong suspicion that it is true. Let me explain this a bit.

Here is a formulation that I find quite instructive: a cocongruence on a group $U$ is a way of putting equivalence relations on the hom-sets $hom(U,G)$ for all groups $G$, such that
- homomorphisms $h : G \to G'$ preserve it
- they are compatible with limits $G = \lim_i G_i$

So we have an equivalence relation $f \sim g$ for homomorphisms $f,g : U \to G$, such that  $f \sim g \implies hf \sim hg$ for $h : G \to G'$, and when $G = \lim_i G_i$, then $f \sim g$ iff $p_i f \sim p_i g$ for all $i$.

We also need that the equivalence relation on $hom(U,-)$ is representable, but in many cases this follows because of the limit compatibility. (The only thing left is accessibility, right?)

It is effective when there is a subgroup $V \subseteq U$ such that the equivalence relation is given by $f \sim g \iff f|_V = g|_V$.

(All that holds similarly for general categories, but I find it instructive to write it down in this special case.)

The special case $U = \mathbb{Z}$ means: on the underlying set of any group $G$ we have a natural equivalence relation (we could call it "universal") which is limit-compatible, and we need to find $n \in \mathbb{Z}$ such that it is given by $a \sim b \iff a^n = b^n$. This is an interesting classification result (if true).

And here is why I think it is true: I would literally fall off my chair if somebody writes down an equivalence relation (with the mentioned properties) that does not have this form (for general groups). Yes, this is no proof.

What I find remarkable is that this does not seem to be trivial at all for **Ab**, but your (almost formal) implication `preadditive_kernels_normal_imply_effective_congruences` handles this case automatically. Maybe here my POV of ignoring the representable object is not ideal. But still, I think we can use $\mathbb{Q} / \mathbb{Z}$ to find the number $n$, alongside with the lemma that (in the abelian case!) the equivalence relation is actually a subgroup.

From this we get some global $n$ such that $a \sim b \iff a^n = b^n$ when $a$ and $b$ commute. But for non-abelian groups, I have no clue (yet). Since **Grp** is mono-regular, one can show that the relation descends to subgroups. In particular, it suffices to look at 2-generated groups, at least in theory. Free groups are of no use since here powers cancel.

Random remark: We are proving here that **Grp** does not satisfy the second half of the definition of being Barr-coexact. But we already know that it is not coregular anyway (the first half of the definition).

Question: I just saw that you have proven that **CRing** does not have effective cocongruences (using `pretopos_balanced`), which I find quite surprising. How does the counterexample look like in the description above?

dschepler:
Yes, that's precisely the line of thought I was going along. One thought which occurs is that the equivalence relation of conjugacy is preserved by homomorphisms. However, it's not representable (at least I think it isn't, else you'd essentially get a counterexample to **Grp** being epi-regular). Another thought: Any proof is going to have to take into account the cotransitivity map in some way. For example, the relation $(a b^{-1})^2 = 1$ is certainly reflexive and symmetric; but it's not transitive, as you can see for example in a dihedral group with $b$ being a reflection and $a,c$ being arbitrary rotations.

As for the example in **CRing**, I think the counterexample is a corelation on $\mathbb{Z} \times \mathbb{Z}$. I could simplify it a bit by using $\mathbb{Z}[1/2]$ instead of $\mathbb{Q}$. Here $\mathbb{Z} \times \mathbb{Z}$ represents the functor taking a ring to its idempotents; and the equivalence relation represented by $\mathbb{Z} \times \mathbb{Z}[1/2] \times \mathbb{Z}[1/2] \times \mathbb{Z}$ is that two idempotents $e_1, e_2$ are equivalent if and only if 2 is invertible in the rings $e_1 (1-e_2) R$ and $(1-e_1) e_2 R$. I guess the picture under **Spec** makes it clearer why this is an equivalence relation, involving the cross terms both missing the point $\langle 2\rangle$. Maybe a similar example using $\mathbb{C}[t]$ and $\mathbb{C}[t,1/t]$ would make an example that's easier to think about geometrically, now with two maps to $\mathbb{A}^1+\mathbb{A}^1$ being equivalent if the projections to $\mathbb{A}^1$ are equal and if the cross terms both miss 0. Except algebraically that complicates the picture because now you also have to keep track of elements of the ring along with idempotents, and maybe now you want $a=b$, and for $e_1 (1-e_2) a$ and $(1-e_1) e_2 a$ to be units in their respective rings.

Anyway, as far as I can tell, all that breaks down completely in noncommutative rings since you no longer have $eR$ forming a sub-rng.

ScriptRaccoon:
Ok. I haven't thought about this long, but why cannot we take $eRe$?

Also, images of two commuting idempotents are still commuting. So their product is still idempotent.

For groups, the conjugation relation is not limit-preserving. I think we can say straight away that the equivalence relation must be some algebraic equation.

dschepler:
Hmm... OK, I guess maybe an idempotent would split a ring into $\begin{bmatrix} eRe & (1-e)Re \\ eR(1-e) & (1-e)R(1-e) \end{bmatrix}$ or something along those lines.

For the comments on groups: yes, certainly in the general case if you take a presentation $(A, R)$ of $X$, then from generators of the kernel of $FA+FA \to E$ you get some algebraic equations in terms of some number of paired variables, which form an equivalence on tuples satisfying the relations $R$.  EDIT: For a slightly different point of view but equivalently, the equations define a partial equivalence relation whose natural domain is the tuples satisfying the relations from $R$.  (Of course that description only works for algebraic categories which are epi-regular.)  The thing is that those equations could nomimally involve cross terms and the proof would have to find some way to detangle the cross terms, either implicitly or explicitly.  The example I have in mind is that $E := \langle a, b \mid b^{-1} a = b a^{-1} \rangle$ as a corelation on $\mathbb{Z}$ is a disguised version of the relation $a^2 = b^2$.  (I guess maybe that's another reason why the proof is so easy in **R-Mod**: you can just move cross terms to the other side. :) )

dschepler:
For another example I have in mind, in the full subcategory of integral (or cancellative) commutative monoids, the equivalence $(a, b) \sim (c, d)$ if and only if $a+d = b+c$ is representable, but not effective.  Of course, there's no reason to believe that category is particularly well-behaved...

dschepler:
Another idea I had for **Ring** which ended up not panning out: defining the congruence relation as having $y-x$ which is divisible. But then I found that divisibility structure isn't necessarily unique, for example when applied on
```
[ Z Q/Z ]
[ 0 Z   ]
```
so that's likely to have issues with not respecting equalizers, or with not giving an epimorphism to the representing object if adjoining a concrete divisibility structure such as an abelian group homomorphism $\mathbb{Q} \to (R, +)$ with $1 \mapsto y-x$.

dschepler:
I wonder if there might be group theorists at MO who might have ideas. Especially if we give the concrete versions in terms of systems of equations in $2 \times \kappa$ variables which induce a partial equivalence relation on $G^\kappa$ for each $G$ (including $F(\kappa)$ so that can also be expressed as the existence of formal implications of symmetry and transitivity in the system). And the question becomes: is that always equivalent to a system of the form $f_i(\vec x) = e, f_i(\vec y) = e, g_j(\vec x) = g_j(\vec y)$.

ScriptRaccoon:
Funny coincidence, since I also wanted to ask on MO right now, but I wanted to coordinate this with you first. Personally, I would like to ask the question about **Z** first, with the reformulation I gave above. But you can go ahead... 

Just to reiterate: I would like to avoid working with words in free groups at all costs, and rather try to use the functorial POV.

PS: I also made several attempts to get an answer from Google Gemini, but they failed.

dschepler:
Sure, I would be fine with starting with asking about the simple case first.  (Feel free to add some of my non-counterexamples if you'd like to use them to illustrate some of the complexities of the question, e.g. $a b^{-1} = a^{-1} b$ as a disguised version of the $n=2$ case, or $(a b^{-1})^2 = e$ as a case where reflexivity and symmetry hold but not transitivity. Another example with reflexivity and symmetry hold but transitivity fails comes from $S_3$ as a corelation on $C_2$, with the maps to (1 2) and (2 3). That has presentation $\langle x, y \mid x^2=1, y^2=1, xyx=yxy \rangle$. But that fails transitivity for example in $S_4$ with elements (1 2), (2 3), (3 4). EDIT: Though that's not an example of the simplified case of course.)

dschepler:
It's funny enough you mentioned Google Gemini - I had my first session with it on a mathematical question yesterday, regarding my question about whether a divisibility structure on an element of the additive group of a ring is necessarily unique. It had several false starts, where I had to point out a flaw in the reasoning - including some first answers where it thought the result was true - until eventually it reminded me of the construction of adjoining a unit to the rng on Q/Z with zero multiplication.

ScriptRaccoon:
I made another attempt with Google Gemini (Pro). This time, the proof looks good. I will spend more time, but for now I cannot find a mistake. (Click for a larger version.)

proof<br />

About that "subgroup theorem for amalgamated free products": probably the name is a hallucination, but the statement appears to be true based on the usual element structure.

Relevant: https://www.researchgate.net/publication/266843789_Subgroups_of_amalgamated_free_products

dschepler:
I wonder if that proof simplifies in any significant way for the special case $G = \mathbb{Z}$, to give some insight into what the proof is doing in that situation. I guess at least $D,X,Y$ become cyclic subgroups.

dschepler:
Also, unless I'm missing it somewhere, it doesn't use cosymmetry. So, that would imply for example that any representable functor giving a preorder on each group in fact gives an equivalence relation on each group - which I guess isn't that surprising.

ScriptRaccoon:
Here is my "human" proof that cocongruences in **Grp** are effective, which is just a slight variation of Gemini's proof. The vague citation of the "subgroup theorem for amalagamted free products" is replaced by a direct argument using the well-known description of the elements of a pushout.

[cocongruences_grp.pdf](https://github.com/user-attachments/files/27088909/cocongruences_grp.pdf) (*outdated, see next comment*)

If you have verified it, I can add a commit which adds it (pdf + tex + `Grp.sql`).

Ah and yes, cosymmetry is not needed.

ScriptRaccoon:
I have improved the proof in many ways and also generalized it. You can have a look at the commit I just added.

I find this result quite beautiful!

So only **Ring** is missing to make (current) unit test work (we can always change that test if the challenge is too big!). But tbh I also would like to know the answer for **Mon**.

databases/catdat/data/002_category-properties/100_related-category-properties.sql
```diff
@@ -291,9 +291,15 @@ VALUES
 ('quotients of congruences', 'coequalizers'),
 ('quotients of congruences', 'cokernels'),
 ('quotients of congruences', 'regular'),
+('quotients of congruences', 'effective congruences'),
 ('coquotients of cocongruences', 'equalizers'),
 ('coquotients of cocongruences', 'kernels'),
 ('coquotients of cocongruences', 'coregular'),
+('coquotients of cocongruences', 'effective cocongruences'),
+('effective congruences', 'normal'),
```
ScriptRaccoon, line 301: Is there a reason you didn't pick "mono-regular" here?

databases/catdat/data/003_category-property-assignments/FS.sql
```diff
@@ -65,6 +65,12 @@ VALUES
 	TRUE,
 	'The empty set and a singleton give a multi-terminal object.'
 ),
+(
+	'FS',
+	'effective congruences',
+	TRUE,
+	'In fact, for any congruence $E \rightrightarrows X$ in $\mathbf{FS}$, we must have the reflexivity morphism $X \to E$ surjective. That implies $E$ is the kernel pair of $\mathrm{id}_X$.'
+),
```
ScriptRaccoon: This looks like it should be handled by an implication.

databases/catdat/data/003_category-property-assignments/PMet.sql
```diff
@@ -138,4 +138,10 @@ VALUES
 	'If $(N,z,s)$ is a natural numbers object in $\mathbf{PMet}$, then
 	$$1 \xrightarrow{z} N \xleftarrow{s} N$$
 	is a coproduct cocone by <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a>, Part A, Lemma 2.5.5. Since there is a map $1 \to N$, we have $N \neq \varnothing$. However, the coproduct of two non-empty pseudo-metric spaces does not exist, see <a href="https://math.stackexchange.com/questions/1778408" target="_blank">MSE/1778408</a>.'
+),
+(
+	'PMet',
+	'effective cocongruences',
+	FALSE,
+	'The proof is similar to the one for <a href="Top">$\mathbf{Top}$</a>: Use the two-point space with the zero metric, which represents the functor taking a pseudo-metric space to the pairs of points with $d(x,y) = 0$. In this case, once you conclude $Z = \varnothing$, the map $h : Z \to 1$ does not have any cokernel pair, since that would have to be a coproduct $1+1$.'
```
ScriptRaccoon: Fix link:

```suggestion
	'The proof is similar to the one for <a href="/category/Top">$\mathbf{Top}$</a>: Use the two-point space with the zero metric, which represents the functor taking a pseudo-metric space to the pairs of points with $d(x,y) = 0$. In this case, once you conclude $Z = \varnothing$, the map $h : Z \to 1$ does not have any cokernel pair, since that would have to be a coproduct $1+1$.'
```

databases/catdat/data/003_category-property-assignments/Set.sql
```diff
@@ -29,6 +29,12 @@ VALUES
 	TRUE,
 	'Use the empty algebraic theory.'
 ),
+(
+	'Set',
+	'effective cocongruences',
+	TRUE,
+	'Since the contravariant powerset functor $\mathbf{Set}^{\mathrm{op}} \to \mathbf{Set}$ is monadic, we conclude that $\mathbf{Set}^{\mathrm{op}}$ is Barr-exact and therefore has effective congruences.'
+),
```
ScriptRaccoon:
```suggestion
(
  -- TODO: rework this when Barr-exact is added to the db
	'Set',
	'effective cocongruences',
	TRUE,
	'Since the contravariant powerset functor $\mathbf{Set}^{\mathrm{op}} \to \mathbf{Set}$ is monadic, we conclude that $\mathbf{Set}^{\mathrm{op}}$ is Barr-exact and therefore has effective congruences.'
),
```

databases/catdat/data/003_category-property-assignments/Set_c.sql
```diff
@@ -77,6 +77,18 @@ VALUES
...
+	'effective congruences',
+	TRUE,
+	'For a countable kernel pair $E$ of $h : X \to Z$ where $X$ is countable, replacing $Z$ with the image of $h$ gives the same kernel pair.'
+),
+(
+	'Set_c',
+	'effective cocongruences',
+	TRUE,
+	'For a countable cokernel pair $E$ of $h : Z \to X$ where $X$ is countable, replacing $h$ with the inclusion map of its image gives the same cokernel pair.'
+),
```
ScriptRaccoon, line 85: Can we extract a lemma here? cf. FinGrp and FinSet.

databases/catdat/data/003_category-property-assignments/Set_f.sql
```diff
@@ -41,6 +41,18 @@ VALUES
 	TRUE,
 	'A congruence on a set $X$ in $\mathbf{Set}_\mathrm{f}$ is the same as an equivalence relation $R$ on $X$ whose equivalence classes are finite. In that case, the usual quotient map $p : X \to X/R$ is finite-to-one. Moreover, if $h : X/R \to Y$ is a map such that $h \circ p : X \to Y$ is finite-to-one, then $h$ is finite-to-one as well because $h^*(\{y\}) \subseteq p^*((h \circ p)^*(\{y\}))$ for all $y \in Y$. Therefore, $p$ is also the quotient in $\mathbf{Set}_\mathrm{f}$.'
 ),
+(
+	'Set_f',
+	'effective congruences',
+	TRUE,
+	'If $E \rightrightarrows X$ is the kernel pair of $h : X \to Z$ and both maps $E \to X$ are finite-to-one, then that means the equivalence classes of $E$ are finite. Thus, necessarily $h$ was finite-to-one also.'
```
ScriptRaccoon:
```suggestion
	'If $E \rightrightarrows X$ is the kernel pair of $h : X \to Z$ in $\mathbf{Set}$ and both maps $E \to X$ are finite-to-one, then that means the equivalence classes of $E$ are finite. Thus, necessarily $h$ was finite-to-one also.'
```

databases/catdat/data/003_category-property-assignments/Set_pointed.sql
```diff
@@ -67,6 +67,12 @@ VALUES
 	TRUE,
 	'The coproduct (wedge sum) of a family of pointed sets $(X_i)_{i \in I}$ can be realized as the subset of $\prod_{i \in I} X_i$ consisting of those tuples $x$ such that $x_i = 0$ for all but (at most) one index.'
 ),
+(
+	'Set*',
+	'effective cocongruences',
+	TRUE,
+	'We have that $\mathbf{Set}_*^{\mathrm{op}}$ is a slice category of $\mathbf{Set}^{\mathrm{op}}$, which in turn is monadic over $\mathbf{Set}$. Therefore, $\mathbf{Set}_*^{\mathrm{op}}$ is Barr-exact, and in particular it has effective congruences.'
+),
```
ScriptRaccoon, line 76:
```suggestion
(
   -- TODO: rework this when Barr-exact is added
	'Set*',
	'effective cocongruences',
	TRUE,
	'We have that $\mathbf{Set}_*^{\mathrm{op}}$ is a slice category of $\mathbf{Set}^{\mathrm{op}}$, which in turn is monadic over $\mathbf{Set}$. Therefore, $\mathbf{Set}_*^{\mathrm{op}}$ is Barr-exact, and in particular it has effective congruences.'
),
```

databases/catdat/data/004_category-implications/001_limits-colimits-existence-implications.sql
```diff
@@ -69,6 +69,13 @@ VALUES
 	'For any congruence $E$ on an object $X$ of a preadditive category, let $E_0$ be the kernel of $p_2 : E \to X$.  The restriction of $p_1$ to $E_0$ is a monomorphism.  We can then see that $E$ must be the pullback of $p_1 - p_2 : E \to X$ and $E_0 \hookrightarrow X$.  Then the cokernel of $E_0 \hookrightarrow X$ is a quotient of $E$.',
 	FALSE
 ),
+(
+	'thin_effective_congruences',
+	'["thin"]',
```
ScriptRaccoon: We need less than `thin`, cf. my remark to FS.sql.

And this result is subsumed by `reflexive_pair_trivial` below. So let's remove it.

databases/catdat/data/004_category-implications/005_additional-structure-implications.sql
```diff
@@ -27,6 +27,13 @@ VALUES
 	'This is trivial.',
 	FALSE
 ),
+(
+	'preadditive_kernels_normal_imply_effective_congruences',
+	'["preadditive", "kernels", "normal"]',
+	'["effective congruences"]',
+	'Let $f, g : E \rightrightarrows X$ be a congruence. Then let $E_0$ be the kernel of $g$. We see that $f$ restricted to $E_0$ is a monomorphism $E_0 \hookrightarrow X$. If this is the kernel of a morphism $h : X \to Y$, then $E$ is the kernel pair of $h$.',
```
ScriptRaccoon: Can you add the proof that it is the kernel pair?

databases/catdat/data/004_category-implications/008_topos-theory-implications.sql
```diff
@@ -121,7 +121,7 @@ VALUES
 (
 	'topos_consequence',
 	'["elementary topos"]',
-	'["finitely cocomplete", "disjoint finite coproducts", "epi-regular"]',
+	'["finitely cocomplete", "disjoint finite coproducts", "epi-regular", "effective congruences"]',
```
ScriptRaccoon, line 124: Is it correct that no new source has been added for this?

databases/catdat/data/003_category-property-assignments/Set.sql
dschepler: Actually, this one is probably redundant now that we have the implication that this is true for any elementary topos - and so is the one for FinSet.  (And I think maybe the hypotheses were weak enough that the explicit entry for the category of countable sets is redundant too.)  I'll check this tomorrow.

databases/catdat/data/004_category-implications/008_topos-theory-implications.sql
dschepler: That was actually one of the things I wanted to come back to and get a reference for. Though I don't have the Mac Lane & Moerdijk book to look up where it is in that (which I'm sure it must be).

Of course, constructing a morphism it's effective for is easy enough: take the subobject of $X\times X$, take its characteristic morphism $X \times X \to \Omega$, and transpose to a morphism $[\cdot]_E : X \to \Omega^X$ (either way, it doesn't matter). Then that's a generalization of the equivalence class function in the case of **Set**. On the other hand, proving it works takes more work, especially the part of proving you get a commutative square (corresponding to $(x,y)\in E \rightarrow [x]_E = [y]_E$). So yes, it would be better just to have a reference.

databases/catdat/data/003_category-property-assignments/FS.sql
ScriptRaccoon: Yes you can just remove this assignment, it is handled by `dual_reflexive_pair_trivial`.

databases/catdat/data/004_category-implications/008_topos-theory-implications.sql
ScriptRaccoon: I gave a reference [here](https://github.com/ScriptRaccoon/CatDat/pull/103/changes#r3106928439) already. It is Part A, Proposition 2.4.1 in the Elephant. For some reason I cannot find it in Mac Lane - Moerdijk. In that book I found the theorem that every epimorphism is [effective](https://ncatlab.org/nlab/show/effective+epimorphism) (Theorem IV.7.8 on p. 197), is this sufficient?

databases/catdat/data/003_category-property-assignments/Set_c.sql
dschepler: I was thinking the next step could be to add property "has strict quotients of congruences" and its dual. This would be nominally "the coequalizer exists and that particular coequalizer map also induces a cartesian square" - but then there's the easy implication to add that it's equivalent to "has quotients of congruences" and "has effective congruences". Then, in cases like this where I'm doing the replacement, I could revise them to be assignments to the "strict quotients" property.

I was also thinking we could stand to be more explicit about the relations between having quotients / effectiveness in a full subcategory vs. the larger category.

databases/catdat/data/002_category-properties/100_related-category-properties.sql
dschepler: I was thinking of the comparison between "every congruence - which is a class of jointly monic pairs - is a kernel pair" and "every monomorphism is a kernel". I guess indeed you could also make a comparison between "every congruence is an 'equalizer pair' of a morphism" and "every monomorphism is an equalizer of a pair of morphisms" though that seems slightly more tenuous. It does seem worth adding, though.

dschepler:
For some minor comments on the **CRing** case regarding the counterexample there:

Geometrically under the image of **Spec**, the equivalence relation is that (2) maps to the same component of $\mathbf{Spec}(\mathbb{Z}) \sqcup \mathbf{Spec}(\mathbb{Z})$ under both maps.

The purely algebraic proof that it works as an equivalence relation: $e_1 (1-e_3) = e_1 e_2 (1-e_3) + e_1 (1-e_2) (1-e_3)$, so we can take $e_1$ times the inverse of 2 in $e_2 (1-e_3) R$ plus $(1-e_3)$ times the inverse of 2 in $e_1 (1-e_2) R$ as an inverse of 2 in $e_1 (1-e_3) R$; and similarly for $(1-e_1) e_3 R$.

dschepler:
Still no progress on the case of **Ring**. Time to see if anybody on MO has any ideas?

I also had the idea of trying to create a counterexample in **CMon** based on the counterexample in **CRing** with $\mathbb{Z}[t] \hookrightarrow \mathbb{Z}[t,1/t]$. That doesn't quite seem to be working out for the conditions involving $1-e$. On the other hand, maybe I should have expected such attempts wouldn't be that easy, based on the fact that **CMon** is not even codistributive, much less coextensive.

ScriptRaccoon:
> For some minor comments on the **CRing** case regarding the counterexample there:
> 
> Geometrically under the image of **Spec**, the equivalence relation is that (2) maps to the same component of Spec ( Z ) ⊔ Spec ( Z ) under both maps.
> 
> The purely algebraic proof that it works as an equivalence relation: e 1 ( 1 − e 3 ) = e 1 e 2 ( 1 − e 3 ) + e 1 ( 1 − e 2 ) ( 1 − e 3 ) , so we can take e 1 times the inverse of 2 in e 2 ( 1 − e 3 ) R plus ( 1 − e 3 ) times the inverse of 2 in e 1 ( 1 − e 2 ) R as an inverse of 2 in e 1 ( 1 − e 3 ) R ; and similarly for ( 1 − e 1 ) e 3 R .

I don't have enough context to understand what you are writing here (which is a general problem in our communication, I must say). "both maps" -> which maps? Can you please start the example from scratch?

> Still no progress on the case of Ring. Time to see if anybody on MO has any ideas?

Yes, you can ask. 👍🏼 To be honest, I have no idea for this category.

dschepler:
https://mathoverflow.net/questions/510744/does-the-category-of-rings-have-effective-cocongruences

ScriptRaccoon:
While we are waiting for MO, maybe there is a chance to handle some of the other categories? 

Unknown categories for "effective congruences":
- Ban*
- PMet*

Unknown categories for "effective cocongruences":
- Alg
- Ban
- CMon
- FinGrp*
- Haus*
- LRS
- Met_c
- Met
- Met_oo
- Mon
- Pos*
- Ring
- Rng
- Sch
- Cat

*These categories look doable.

databases/catdat/data/003_category-property-assignments/FinGrp.sql
```diff
@@ -65,6 +65,12 @@ VALUES
 	TRUE,
 	'The proof works exactly as for the <a href="/category/FinSet">category of finite sets</a>.'
 ),
+(
+	'FinGrp',
+	'effective congruences',
+	TRUE,
+	'For a kernel pair $E$ of $h : X \to Z$ where $E$ and $X$ are finite groups, replacing $Z$ with the image of $h$ gives the same kernel pair.'
```
ScriptRaccoon: Please elaborate this proof. Why are you starting with a kernel pair? We should start with a congruence?
I assume you are using that **Grp** has effective congruences? If so, I suggest to make this explicit.

databases/catdat/data/003_category-property-assignments/FreeAb.sql
```diff
@@ -61,6 +61,12 @@ VALUES
 	(1) If $f : A \to B$ is surjective, it is the coequalizer of $A \times_B A \rightrightarrows A$ in $\mathbf{Ab}$. Since $A \times_B A$ is free abelian, $f$ is also an coequalizer in $\mathbf{FreeAb}$.<br>
 	(2) If $f : A \to B$ is a regular epimorphism in $\mathbf{FreeAb}$, consider the factorization $f = i \circ p$ as above. Since $f$ is an extremal epimorphism, $i$ must be an isomorphism, so that $f$ is surjective.'
 ),
+(
+	'FreeAb',
+	'effective cocongruences',
+	TRUE,
+	'Since $\mathbf{Ab}$ is abelian, it has effective cocongruences. Now, suppose a cocongruence $X \rightrightarrows E$ is the cokernel pair of $h : Z \to X$, where $X$ and $E$ are free abelian groups. If we find a surjective map $g : F \to Z$ where $F$ is a free abelian group, then $h \circ g$ has the same cokernel pair as $h$, and $h \circ g$ is a morphism in $\mathbf{FreeAb}$.'
```
ScriptRaccoon: Are you using `preadditive_kernels_normal_imply_effective_congruences` in the first sentence? In any case, please add a reference.

ScriptRaccoon: I assume this is the intention?

```suggestion
	'Since $\mathbf{Ab}$ is abelian, it has effective cocongruences. Now, suppose a cocongruence $X \rightrightarrows E$ in $\mathbf{FreeAb}$ is the cokernel pair in $\mathbf{Ab}$ of $h : Z \to X$. If we find a surjective map $g : F \to Z$ where $F$ is a free abelian group, then $h \circ g$ has the same cokernel pair as $h$, and $h \circ g$ is a morphism in $\mathbf{FreeAb}$.'
```

But why is every cocongruence in **FreeAb** also one in **Ab**? We don't have pushouts. So we just have the functorial definition. But then we have less test objects.

Same problem with **FinGrp**.

databases/catdat/data/003_category-property-assignments/Meas.sql
```diff
@@ -100,4 +100,10 @@ VALUES
 	'cofiltered-limit-stable epimorphisms',
 	FALSE,
 	'We already know that $\mathbf{Set}$ does not have this property. Now apply the contrapositive of the dual of <a href="/lemma/filtered-monos">this lemma</a> to the functor $\mathbf{Set} \to \mathbf{Meas}$ which equips a set with the trivial $\sigma$-algebra.'
+),
+(
+	'Meas',
+	'effective cocongruences',
+	FALSE,
+	'The proof is similar to the one for <a href="/category/Top">$\mathbf{Top}$</a>: Use the trivial $\sigma$-algebra on a two-point space.'
```
ScriptRaccoon:
```suggestion
	'The proof is similar to the one for <a href="/category/Top">$\mathbf{Top}$</a>: Use the trivial $\sigma$-algebra on a two-point set.'
```

(without the sigma-algebra is not a "space" yet)

databases/catdat/data/003_category-property-assignments/Met.sql
ScriptRaccoon, line 1: This reminds me of the situation that many categories do not have a subobject classifier simply because not every monomorphism is regular. So it makes sense to define regular subobject classifiers.

subobject classifier exists <==> mono-regular + regular subobject classifier exists

If every congruence is effective, this says that at least certain monomorphisms are regular. But not too much either.

databases/catdat/data/003_category-property-assignments/PMet.sql
```diff
@@ -138,4 +138,10 @@ VALUES
 	'If $(N,z,s)$ is a natural numbers object in $\mathbf{PMet}$, then
 	$$1 \xrightarrow{z} N \xleftarrow{s} N$$
 	is a coproduct cocone by <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a>, Part A, Lemma 2.5.5. Since there is a map $1 \to N$, we have $N \neq \varnothing$. However, the coproduct of two non-empty pseudo-metric spaces does not exist, see <a href="https://math.stackexchange.com/questions/1778408" target="_blank">MSE/1778408</a>.'
+),
+(
+	'PMet',
+	'effective cocongruences',
+	FALSE,
+	'The proof is similar to the one for <a href="/category/Top">$\mathbf{Top}$</a>: Use the two-point space with the zero metric, which represents the functor taking a pseudo-metric space to the pairs of points with $d(x,y) = 0$. In this case, once you conclude $Z = \varnothing$, the map $h : Z \to 1$ does not have any cokernel pair, since that would have to be a coproduct $1+1$.'
```
ScriptRaccoon:
```suggestion
	'The proof is similar to the one for <a href="/category/Top">$\mathbf{Top}$</a>: Use the two-point space with the zero metric, which represents the functor taking a pseudo-metric space to the pairs of points with $d(x,y) = 0$. In this case, once you conclude $Z = \varnothing$, the map $h : Z \to 1$ does not have any cokernel pair, since that would have to be a coproduct $1+1$, which does not exist.'
```

databases/catdat/data/003_category-property-assignments/Set_c.sql
```diff
@@ -77,6 +77,12 @@ VALUES
 	TRUE,
 	'This is because $\{0,1\}$ is a subobject classifier in $\mathbf{Set}$, which is countable, and the monomorphisms coincide.'
 ),
+(
+	'Set_c',
+	'effective congruences',
+	TRUE,
+	'For a kernel pair $E$ of $h : X \to Z$ where $E$ and $X$ are countable, replacing $Z$ with the image of $h$ gives the same kernel pair.'
```
ScriptRaccoon: Here I have the same comments as for **FreeAb**.

databases/catdat/data/003_category-property-assignments/Set_f.sql
```diff
@@ -41,6 +41,18 @@ VALUES
 	TRUE,
 	'A congruence on a set $X$ in $\mathbf{Set}_\mathrm{f}$ is the same as an equivalence relation $R$ on $X$ whose equivalence classes are finite. In that case, the usual quotient map $p : X \to X/R$ is finite-to-one. Moreover, if $h : X/R \to Y$ is a map such that $h \circ p : X \to Y$ is finite-to-one, then $h$ is finite-to-one as well because $h^*(\{y\}) \subseteq p^*((h \circ p)^*(\{y\}))$ for all $y \in Y$. Therefore, $p$ is also the quotient in $\mathbf{Set}_\mathrm{f}$.'
 ),
+(
+	'Set_f',
+	'effective congruences',
+	TRUE,
+	'If $E \rightrightarrows X$ is the kernel pair of $h : X \to Z$ in $\mathbf{Set}$ and both maps $E \to X$ are finite-to-one, then that means the equivalence classes of $E$ are finite. Thus, necessarily $h$ was finite-to-one also.'
```
ScriptRaccoon: You are repeating parts of the proof of `quotients of congruences` above. Let's remove the redundancy. SInce #134 you can safely reference previous proofs.

databases/catdat/data/003_category-property-assignments/Top_pointed.sql
```diff
@@ -168,7 +168,16 @@ VALUES
 	'cofiltered-limit-stable epimorphisms',
 	FALSE,
 	'We already know that $\mathbf{Set}_*$ does not have this property. Now apply the contrapositive of the dual of <a href="/lemma/filtered-monos">this lemma</a> to the functor $\mathbf{Set}_* \to \mathbf{Top}_*$ that equips a pointed set with the indiscrete topology.'
+),
+(
+	'Top*',
+	'effective congruences',
+	FALSE,
+	'Suppose that $\mathbf{Top}_*$ had effective congruences. Then for any congruence $E \rightrightarrows X$ in $\mathbf{Top}$, we can expand it to a congruence $E + \{*\} \rightrightarrows X + \{*\}$ in $\mathbf{Top}_*$. If $E + \{*\}$ is the kernel pair of $h : X + \{*\} \to Z$, then $E$ is the kernel pair of $h$ restricted to $X$. This contradicts the fact that $\mathbf{Top}$ does not have effective congruences.'
```
ScriptRaccoon: For transitivity of $E + \ast \rightrightarrows X + \ast$ we need that the canonical morphism

$(E \times_X E) + \ast \longrightarrow (E + \ast) \times_{X + \ast} (E + \ast)$

is an isomorphism, right? This is not formal.

databases/catdat/data/003_category-property-assignments/Top_pointed.sql
```diff
@@ -168,7 +168,16 @@ VALUES
...
+),
+(
+	'Top*',
+	'effective cocongruences',
+	FALSE,
+	'Consider the pointed topological space $I := \{ *, a, b \}$ with topology $\{ \varnothing, \{ * \}, \{ a, b \}, \{ *, a, b \} \}$. This represents the functor which sends a pointed topological space $X$ to the pairs of indistinguishable points of $X$. Therefore, we get a cocongruence $\{ *, a \} \rightrightarrows I$ on the discrete space $\{ *, a \}$, where the maps are the two possible pointed functions. However, this cannot be effective: if we have $h : Z \to \{ *, a \}$ which equalizes the cocongruence, then $Z$ must be the singleton pointed space. But that means the cokernel pair of $h$ is the discrete space on $\{ *, a, b \}$.'
```
ScriptRaccoon: There are three pointed functions $\\{  *, a  \\} \to I$, not two.

ScriptRaccoon: > if we have $h : Z \to \{ *, a \}$ which equalizes the cocongruence, then $Z$ must be the singleton pointed space.

This is not correct. We can only say that the image of $Z$ is a singleton.

ScriptRaccoon: For this and related proofs where you clearly reduce the proof to another category (here: **Top**), it is best to add a lemma.

databases/catdat/data/003_category-property-assignments/Z.sql
ScriptRaccoon, line 1: Let's just say that this property is inherited from **Set**.

databases/catdat/data/002_category-properties/001_limits-colimits-existence.sql
```diff
@@ -343,6 +343,38 @@ VALUES
...
+	'effective congruences',
+	'has',
+	'A congruence $f, g : E \rightrightarrows X$ (see definition <a href="/category-property/quotients_of_congruences">here</a>) is <i>effective</i> if it is the kernel pair of some morphism, i.e. if there is a morphism $h : X \to Y$ such that we have a cartesian square
+	$$
+	\begin{CD}
+	E @> f >> X \\
+	@V g VV @V h VV \\
+	X @> h >> Y.
+	\end{CD}
+	$$
```
ScriptRaccoon, line 356: Suggestion: put the labels outside.

```suggestion
	$$
	\begin{CD}
	E @> f >> X \\
	@V g VV @VV h V \\
	X @>> h > Y.
	\end{CD}
	$$
```

Similarly for effective cocongruences below.

databases/catdat/data/004_category-implications/005_additional-structure-implications.sql
```diff
@@ -27,6 +27,15 @@ VALUES
 	'This is trivial.',
 	FALSE
 ),
+(
+	'preadditive_kernels_normal_imply_effective_congruences',
+	'["preadditive", "kernels", "normal"]',
+	'["effective congruences"]',
+	'Let $f, g : E \rightrightarrows X$ be a congruence. Then let $E_0$ be the kernel of $g$. We see that $f$ restricted to $E_0$ is a monomorphism $E_0 \hookrightarrow X$. Let $f |_{E_0}$ be the kernel of a morphism $h : X \to Y$. We claim that $E$ is also the kernel pair of $h$.<br>
+	To see this, suppose we have a pair of generalized elements $x_1, x_2 : T \rightrightarrows X$. Then the pair $x_1, x_2$ factors through $E$ if and only if $x_1 - x_2, 0$ does. This is equivalent to the condition that $x_1 - x_2$ factors through $E_0$. That in turn is equivalent to $h \circ (x_1 - x_2) = 0$, which is equivalent to $h \circ x_1 = h \circ x_2$.<br>
+	In particular, applying the forward implications in the case $T := E, x_1 := f, x_2 := g$, we conclude that $h \circ f = h \circ g$, so we get the required commutative diagram. From there, the reverse implications show this diagram is a cartesian square.',
```
ScriptRaccoon:
```suggestion
	In particular, applying the forward implications in the case $T := E$, $x_1 := f$, $x_2 := g$, we conclude that $h \circ f = h \circ g$, so we get the required commutative diagram. From there, the reverse implications show this diagram is a cartesian square.',
```

(will improve line breaks on small viewports)

databases/catdat/data/004_category-implications/007_locally-presentable-implications.sql
```diff
@@ -223,10 +223,17 @@ VALUES
 	'This follows from one of equivalent formulations of multi-algebraic categories.',
 	TRUE
 ),
+(
+	'multi-algebraic_implies_effective_congruences',
+	'["multi-algebraic"]',
+	'["effective congruences"]',
+	'This is Thm. 4.0 in <a href="https://doi.org/10.1007/BF01224953" target="_blank">Yves Diers, Catégories Multialgébriques</a> or <a href="https://translations.thosgood.net/ADM-34-1980-193.pdf" target="_blank">its English translation</a>.',
```
ScriptRaccoon, line 230: just saying that this reference states that this is true, repeats the statement in the proof, but does not give any proof.

databases/catdat/data/004_category-implications/008_topos-theory-implications.sql
```diff
@@ -167,6 +167,29 @@ VALUES
...
+	'["effective cocongruences", "co-Malcev"]',
+	'Suppose we have a coreflexive corelation on $X$, $X+X \overset{p}{\twoheadrightarrow} E \overset{r}{\twoheadrightarrow} X$. Let $Y$ be the equalizer of $p\circ i_1, p\circ i_2 : X \to E$. Then by the assumptions $p$ is a regular epimorphism. By regularity, $p$ is the coequalizer of its kernel pair, which can be expressed as the equalizer of $p\circ p_1, p\circ p_2 : (X+X) \times (X+X) \to E$. By distributivity and extensitivity, it is sufficient to calculate the equalizer on each quadrant of $(X+X) \times (X+X)$. On the $(1,1)$ quadrant, this is the equalizer of $p\circ i_1\circ p_1, p\circ i_1\circ p_2$, which is isomorphic to $X$ since $p\circ i_1$ is a split monomorphism. On the $(1,2)$ quadrant, it is the equalizer of $p\circ i_1\circ p_1, p\circ i_2\circ p_2$. Since $r$ is a common section of $p\circ i_1$ and $p\circ i_2$, any generalized element of this equalizer has equal components; thus, the equalizer is isomorphic to the equalizer $Y$ of $p\circ i_1, p\circ i_2$. Similarly, on the $(2,1)$ quadrant, it is isomorphic to $Y$, and on the $(2,2)$ quadrant, it is isomorphic to $X$.<br>
```
ScriptRaccoon: I was confused for a minute because of the notation $p_1,p_2$ for the product projections which have nothing to do with $p$. (In other contexts, you have used $p_1,p_2$ to denote the components of $p$.) Maybe write $\pi_1, \pi_2$ here and also define them?

ScriptRaccoon: I suggest to replace
$p\circ p_1, p\circ p_2 : (X+X) \times (X+X) \to E$
with
$p\circ p_1, p\circ p_2 : (X+X) \times (X+X) \rightrightarrows E$

ScriptRaccoon: > Since  $r$ is a common section

It is a common retraction.

databases/catdat/data/004_category-implications/008_topos-theory-implications.sql
```diff
@@ -167,6 +167,29 @@ VALUES
...
+	FALSE
+),
+(
+	'pretopos_balanced',
+	'["effective congruences", "extensive"]',
+	'["balanced"]',
+	'Let $i : Y \to X$ be both a monomorphism and an epimorphism. Now define a congruence $f, g : X + Y + Y + X \rightrightarrows X+X$ acting as $i_1,i_1$ on the first copy of $X$; $i_1\circ i, i_2\circ i$ on the first copy of $Y$; $i_2\circ i, i_1\circ i$ on the second copy of $Y$; and $i_2\circ i_2$ on the second copy of $X$. We use extensitivity in showing that $f, g$ are jointly monomorphic, and again in proving transitivity. Now suppose this is the kernel pair of a morphism $h : X \to Z$. Then consider the map pair $i_2, i_1 : X \to X+X$. We must have $h \circ i_2 \circ i = h \circ i_1 \circ i$ since the pair of maps $i_2\circ i, i_1\circ i$ factor through $E$. Since $i$ is an epimorphism, that implies $h\circ i_2 = h\circ i_1$, so $i_2, i_1$ factor through $E$ as well. By disjointness of coproducts, we can conclude that $i_2, i_1$ factor uniquely through the second copy of $Y$. We thus get a morphism $X \to Y$ which is a left inverse of $i$, showing that $i$ must in fact be an isomorphism.',
```
ScriptRaccoon: $i_2 \circ i_2$ should be $i_2, i_2$.

ScriptRaccoon: Let's replace `map pair $i_2, i_1 : X \to X+X$` with `map pair $i_2, i_1 : X \rightrightarrows X+X$`

ScriptRaccoon: The domain of $h$ should be $X + X$, not $X$.

ScriptRaccoon: Let us rename $i$ to something else, maybe $\alpha$, to make the contrast to the coproduct inclusions big.

ScriptRaccoon: Doesn't this proof shows the stronger statement that `effective congruences + extensive => mono-regular`? 

*EDIT.* I am not 100% sure here; maybe this requires $h$ to be the quotient of the congruence

ScriptRaccoon: Here is how I would write down the proof. I like to work with generalized elements, since for me it makes the proof more easy to understand and also the motivation (=translating the proof from sets) more clear. Furthermore, I don't need to introduce notation for morphisms.

proof<br />

--- 
Notice that ' needs to be written as '' in SQlite, hence all these '' below. They get rendered as '.

```
Let $\alpha : A \to B$ be a monomorphism (and also an epimorphism, which we only need later). Let $B''$ be a copy of $B$, and likewise $A''$ be a copy of $A$. Consider the congruence on $B + B''$ generated by $\alpha(a) \sim \alpha(a)''$ for $a \in A$. Formally, we define $E := B + B'' + A + A''$ and define the two morphisms $f,g : E \rightrightarrows B + B''$ by extending the identity on $B + B''$ and
$$f(a) = \alpha(a),\quad f(a'') = \alpha(a)'',$$
$$g(a) = \alpha(a)'',\quad g(a'') = \alpha(a),$$
on generalized elements. Extensitivity can be used to show that $f,g$ are jointly monomorphic. Clearly, the pair $f,g$ is reflexive and symmetric. For transitivity, one uses again extensivity. By assumption, there is a morphism $h : B + B'' \to C$ such that $f,g$ is the kernel pair of $h$, that is, two generalized elements $x,y \in B + B''$ satisfy $h(x)=h(y)$ satisfy iff $x=f(e)$, $y=g(e)$ for some $e \in E$. In particular, we have $h(\alpha(a)) = h(\alpha(a)'')$ for all $a \in A$. Now assume that $\alpha$ is also an epimorphism. Then we get $h(b) = h(b'')$ for all $b \in B$. Then for all $b \in B$ there is some $e \in E$ with $b = f(e)$ and $b'' = g(e)$. Hence, there is some $a \in A$ with $b = \alpha(a)$ and $b'' = \alpha(a)''$. This shows that $\alpha$ is surjective on generalized elements, i.e. that $\alpha$ is a split epimorphism. Since $\alpha$ is also a monomorphism, we win.
```

dschepler:
I'll take a closer look at the comments this evening. For now, I think among the (easy to trivial) lemmas I'll use to plug the gaps in proofs for subcategories will be: If $f, g : E \rightrightarrows X$ is a congruence which has a quotient $p : X \to X/E$ and which is effective, then in fact we get a cartesian square with $f, g, p, p$. And: If $U : \mathcal{C} \to \mathcal{D}$ is fully faithful, and $U$ applied to a square (not necessarily even a commutative square in the assumptions) gives a cartesian square, then the original square was cartesian as well.

So, for example, the proof that **FinGrp** has effective congruences would be: take the coequalizer in the category of groups, which is finite; then we get a cartesian square in **Grp** involving the quotient; so we also get the required cartesian square in **FinGrp**.

databases/catdat/data/003_category-property-assignments/Z.sql
dschepler: If you just choose a random map $h : F(R) \to Z_R$ for each $R$ which induces a cartesian square, then there's no reason to expect to be able to assemble the $Z_R$ into a functor.  I was using the lemma about being able to use the actual quotients instead of $Z_R$ in order to construct a functor.

databases/catdat/data/003_category-property-assignments/FreeAb.sql
dschepler: You're right on **FreeAb** - I don't have any idea anymore, so I've removed that entry.

dschepler:
I'm out of time for this evening. I think I addressed most of the comments, though at the moment I'm not happy with the expanded wording on the $\mathbf{Top}_*$ example. I think maybe the more general lemma there is: if a base category is extensive, and a coslice category has effective congruences, then the base category also has effective congruences.

dschepler:
Regarding generalized elements, I wonder which of the following notations (treating a general category as "being like" a category of sheaves to some extent or another) you accept:
$X(U)$ as a notation for $\mathrm{Hom}(U, X)$
$f(x) \in Y(U)$ for $f : X \to Y$, $x \in X(U)$ as a notation for $f \circ x$ (I see you used this one in the restatement of the proof)
$\alpha^* x \in X(V)$ for $\alpha : V \to U$, $x \in X(U)$ as a notation for $x \circ \alpha$.

databases/catdat/data/003_category-property-assignments/Ring.sql
```diff
@@ -96,4 +96,10 @@ VALUES
...
 	FALSE,
 	'We know that $\mathbf{CRing}$ does not have this property. Now use the contrapositive of the dual of <a href="/lemma/filtered-monos">this lemma</a> applied to the forgetful functor $\mathbf{CRing} \to \mathbf{Ring}$. It preserves epimorphisms by <a href="https://math.stackexchange.com/questions/5133488" target="_blank">MSE/5133488</a>.'
-);
\ No newline at end of file
+),
+(
+	'Ring',
+	'effective cocongruences',
+	FALSE,
+	'See <a href="https://mathoverflow.net/a/510809" target="_blank">MO/510809</a>.'
```
ScriptRaccoon:
```suggestion
	'See <a href="https://mathoverflow.net/a/510809" target="_blank">MO/510744</a>.'
```

answer ID != thread ID

databases/catdat/data/003_category-property-assignments/Mon.sql
```diff
@@ -100,5 +100,11 @@ VALUES
...
+),
+(
+	'Mon',
+	'effective cocongruences',
+	FALSE,
+	'We adapt the counterexample from <a href="https://mathoverflow.net/a/510809" target="_blank">MO/510809</a> for $\mathbf{Ring}$. Namely, let $X$ be the multiplicative monoid of $\{ 0, 1 \}$, which has presentation $\langle p \mid p^2 = p \rangle$, and let $E$ be the monoid with presentation $\langle p, q \mid p^2 = p, q^2 = q, pq = q, qp = p \rangle$. Then $X$ represents the functor sending a monoid $M$ to its idempotents, and $E$ represents the relation on idempotents $a, b$ of $M$ that $ab = b, ba = a$. This is clearly reflexive and symmetric. For transitivity, suppose $ab = b, ba = a, bc = c, cb = b$. Then $ac = a(bc) = (ab)c = bc = c$ and $ca = c(ba) = (cb)a = ba = a$. Therefore, $E$ is a cocongruence on $X$.<br>
```
ScriptRaccoon: `<a href="https://mathoverflow.net/a/510744" target="_blank">MO/510809</a>`

ScriptRaccoon:
> Regarding generalized elements, I wonder which of the following notations [...] you accept:

all of them! :)

databases/catdat/data/009_lemmas/001_lemmas.sql
```diff
@@ -129,4 +129,29 @@ INSERT INTO lemmas (
...
+    Now if this congruence is the kernel pair of $h : A+X \to Z$ in $A \backslash \mathcal{C}$, then $E$ is the kernel pair of $h \circ i_2 : X \to Z$ in $\mathcal{C}$. Namely, if we have two generalized elements $x_1, x_2 : T \rightrightarrows X$ such that $h \circ i_2 \circ x_1 = h \circ i_2 \circ x_2$, then we can construct a map pair $\mathrm{id}_A + x_1, \mathrm{id}_A + x_2 : A+T \to A+X$ in $A \backslash \mathcal{C}$ with $h \circ (\mathrm{id}_A + x_1) = h \circ (\mathrm{id}_A + x_2)$. Therefore, the pair of maps $\mathrm{id}_A + x_1, \mathrm{id}_A + x_2$ factors through $A+E$, so $x_1, x_2$ factors through $A+E$; and using disjoint coproducts, we may conclude $x_1, x_2$ factors through $E$.'
```
dschepler: It almost looks like this could be generalized to a statement about adjoint pairs (in this case the left adjoint being $X \mapsto A+X$ from $\mathcal{C}$ to $A \backslash \mathcal{C}$ and the right adjoint being the forgetful functor from $A \backslash \mathcal{C}$ to $\mathcal{C}$) -- except that we're also making essential use of the specific form of the left adjoint as a disjoint coproduct.

ScriptRaccoon:
Great to see all the progress here!

Please keep me updated and tell me when you think it is done. Then I will do another review.

I have just merged https://github.com/ScriptRaccoon/CatDat/pull/142. Can you please [rebase](https://git-scm.com/docs/git-rebase) and make the adjustment in `TorsFreeAb.sql` where I already wrote a comment? That is, replace the property in the last assignment with `effective congruences` and adjust the text accordingly.

I haven't thought about if **TorsFreeAb** has effective cocongruences. Maybe it is already deduced automatically. But if not, maybe you can find a proof.

ScriptRaccoon:
Maybe https://github.com/ScriptRaccoon/CatDat/pull/149 has some (minor) interesting consequences for this PR.

... and merge conflicts, sorry!

databases/catdat/data/003_category-property-assignments/Met.sql
```diff
@@ -144,4 +144,18 @@ VALUES
...
+),
+(
+	'Met',
+	'effective cocongruences',
+	FALSE,
+	'We will define a cocongruence on $(0,1)$ where $E := (-1, 0) \cup (0, 1)$ with the standard subspace metric from $\mathbb{R}$, and the two maps $(0, 1) \rightrightarrows E$ are the inclusion map and $x \mapsto -x$. Then for any metric space $X$, the induced relation on non-expansive maps $(0, 1) \to X$ is that $f \sim g$ if and only if $d(f(x), g(y)) \le x+y$ for each $x, y \in (0, 1)$. This is reflexive since $d(f(x), f(y)) \le |x-y| < x+y$, and it is clearly symmetric. For transitivity, suppose $f\sim g$ and $g\sim h$.  Then for any $\varepsilon > 0$, we have $d(f(x), h(y)) \le d(f(x), g(\varepsilon)) + d(g(\varepsilon) + h(y)) \le (x + \varepsilon) + (y + \varepsilon)$. Since this holds for every $\varepsilon > 0$, we conclude $d(f(x), h(y)) \le x+y$.<br>
+	On the other hand, if this cocongruence were effective, then by the dual of <a href="/lemma/effective-congruence-quotients">this result</a>, it would be the cokernel pair of the equalizer of the two inclusion maps. However, that equalizer is empty, so $E$ would have to be a binary copower of $(0,1)$, which does not exist in $\mathbf{Met}$.'
+);
```
dschepler, line 164: The nature of this counterexample, along with my failure to find counterexamples on small finite $X$, leads me to the conjecture that the category of complete metric spaces with non-expansive maps does have effective cocongruences.

dschepler:
> Please keep me updated and tell me when you think it is done. Then I will do another review.

I think it is done and ready for review. (Unless you want me to rebase instead of leaving the merges as is, in which case I would probably have to come back to that tomorrow.)

ScriptRaccoon:
Great!

I will do a detailed review now. I will also make some commits with minor edits if required.

Regarding rebase, I will write a separate comment. That will only be relevant before we merge, so not now.

ScriptRaccoon:
I added a commit with various minor adjustments. Most of them improve the readability of the formulas. Please check these files because I made a bit more there:

- `005_additional-structure-implications.sql` (used a calculation with generalized elements)
- `001_lemmas.sql` (changed the lemma title)
- `FinGrp.sql` (simplification of the proof)

In the proof of `regular_epi-regular_extensive_consequences` I would like to see more details in the paragraph "Since the $X$ pieces are already equalized by the kernel pair ...". Maybe using a bit more formulas would make it more clear what is happening here.

Apart from that, everything looks very good now. Thanks a lot for this massive work! This PR has added a lot of new material, both general results and interesting proofs for specific categories.

As for the merge, I would like to request two things:

- no merge commits
- only a few commits (max. 10)

How I would do it: Make sure that the main branch is uptodate. Then, in the feature branch, execute `git rebase -i main`. Change the order of the commits and [squash them](https://www.geeksforgeeks.org/git/use-of-git-squash-commits/) into each other when they are related.

For a more brutal approach, and I think this is necessary here: Execute `git reset --soft main`. Now, all commits are gone, but the changes are in the staging area. Unstage all files. Now, make a few commits based on topic, and commit all the changes per file. For this PR, the 3 commits could be:

- add properties: effective congruences / effective cocongruences

This commit contains the changes in the files `001_limits-colimits-existence.sql` and `100_related-category-properties.sql`.

- add results on congruences

This commit contains the changes in the files`001_lemmas.sql` and all the files in the folder `/004_category-implications/`.

- decide effective (co)congruences for most categories

This contains all other changes. (It can also be split into several commits, but not too many.)

When no file is left, force push via`git push -f`.

I do not recommend merge commits. For future PRs, I suggest to use `git rebase main` instead.

This will yield a clean commit history in the main branch. Of course we lose some of the history during the development of this PR, but I don't think that any of this is relevant in the future. The results matter, and these are kept, of course.

These remarks apply to all PRs. I wanted to write already the same in #115 (which had 40 commits, which could be grouped to 3 commits as well), but didn't want to overwhelm you with "rules" (as if there weren't already enough).

So this is what I recommend, but if you have strong objections, or need help, please let me know.

databases/catdat/data/003_category-property-assignments/Pos.sql
```diff
@@ -94,4 +94,11 @@ VALUES
...
 	FALSE,
 	'Pick any poset $X$ which has a decreasing sequence of non-empty sets $X = X_0 \supseteq X_1 \supseteq \cdots$ with empty intersection, such as $X_n = \mathbb{N}_{\geq n}$ with the natural ordering. The unique map $X_n \to 1$ is surjective, but their limit $\varnothing \to 1$ is not surjective.'
+),
+(
+	'Pos',
+	'effective cocongruences',
+	FALSE,
+	'Let $X$ be $\mathbb{R}$ with the standard (total) order, and let $E$ be the poset with underlying set $\mathbb{R} \times \{ 0, 1 \}$ and partial order such that $(x, m) \le (y, n)$ if and only if $x < y$ or $(x, m) = (y, n)$. The two maps $\mathbb{R} \rightrightarrows E$ will be $x \mapsto (x, 0)$ and $x \mapsto (x, 1)$ respectively. For any partial order $(\mathbb{P}, \le)$, the induced equivalence relation on the set of order-preserving functions $\mathbb{R} \to \mathbb{P}$ is that $f \sim g$ if and only if $f(x) \le g(y)$ and $g(x) \le f(y)$ whenever $x < y$. This relation is clearly reflexive and symmetric; for transitivity, if $f \sim g$ and $g \sim h$, then whenever $x < y$, we have $f(x) \le g(\frac{x+y}{2}) \le h(y)$ and similarly $h(x) \le g(\frac{x+y}{2}) \le f(y)$, showing that $f \sim h$.<br>
```
ScriptRaccoon: Things like `\mathbb{R}` can be abbreviated to `\IR` after merging #152 . I wonder which PR we should merge first? How to minimize the work?

I think it makes sense to merge #152 first. I can also make the required adjustments here. There will be a merge conflict in `TorsFreeAb.sql` which is easy to handle.

dschepler:
> In the proof of `regular_epi-regular_extensive_consequences` I would like to see more details in the paragraph "Since the X pieces are already equalized by the kernel pair ...". Maybe using a bit more formulas would make it more clear what is happening here.

I've just pushed a change to clarify things in terms of generalized elements.

I wonder if it would also make sense to remove the topos_is_co-Malcev implication which should be redundant now. Maybe it could make sense to move the citation from there to the comment you added at the end of regular_epi-regular_extensive_consequences.

ScriptRaccoon:
Sorry for producing millions of merge conflicts ...

> I wonder if it would also make sense to remove the topos_is_co-Malcev implication which should be redundant now.

Yes, let's remove it.


> Maybe it could make sense to move the citation from there to the comment you added at the end of regular_epi-regular_extensive_consequences.

Yes. Please phrase it as optional information, since the reader already has the proof above.

> I've just pushed a change to clarify things in terms of generalized elements.

Thanks! I will have a look.

*EDIT.* The proof is much better now. 👍🏼

databases/catdat/data/003_category-property-assignments/Pos.sql
dschepler: > Things like `\mathbb{R}` can be abbreviated to `\IR` after merging #152 . I wonder which PR we should merge first? How to minimize the work?
> 
> I think it makes sense to merge #152 first. I can also make the required adjustments here. There will be a merge conflict in `TorsFreeAb.sql` which is easy to handle.

I see you just merged #152 a few minutes ago. I'm done with the other content revision you requested - so you can make the required adjustments if you want without worrying about stepping on my toes. Or, if you prefer, I can merge it and do the adjustments, and then do the requested history squash.

ScriptRaccoon: <s>Ok I will do the adjustments now. This will hopefully make the rebase also easier. Remark: this means the formulas will not render properly for the time being ...</s>

ScriptRaccoon: No wait, this does not make any sense. The process will be much smoother if we first incorporate the main branch with 152 in particular. This is because the macros are optional to use. But for using the macros, their definition is not optional ...

Can you first do* `git reset --soft main` and "collapse" the commits as explained in the other comment? (I don't want to do it since then all the commits will belong to me, but it's your work.)

*Then* I can go ahead and change the formulas.

*I just noticed that this produces quite a lot of files that have not been changed in your PR. This is the problem of merge commits, the history diverges too much ... I hope we don't need to repeat #97

ScriptRaccoon: It seems you managed to do it? :) 

Ah regarding

> Note that ScriptRaccoon <scriptraccoon@gmail.com> contributed many enhancements to
the proofs and presentation, and in particular contributed the proof that Grp has
effective cocongruences.

Maybe it is better to remove the additions for `Grp` then. I can commit this again. I have the TeX file here.

I totally forgot about that proof, isn't this months ago? :D

dschepler: Yes, the squash into 3 commits worked pretty smoothly as described. The only addition I needed to make was: after the `git reset --soft main` it still had all the branch changes staged, so I also needed to `git reset` and then do the selective adds and commits.

ScriptRaccoon: Alright. In case you didn't see my edit:

> Maybe it is better to remove the additions for Grp then. I can commit this again. I have the TeX file here.

dschepler:
I've just done another rebase with the Grp contributions sorted into a separate commit (for which I used --author to credit you).

I also found online the git workflow where you establish a few topics up front, and then for subsequent commits to add to a topic you set a commit message of the form:
fixup! general results
More explanation for reviewers in the meantime before the next rebase

So that you prepare for using "git rebase -i --autosquash main" later.

I think I'll be using that workflow on future PRs.  (And if you want, you can establish a topic or two for commits that you contribute and do the same.)

ScriptRaccoon:
Yes I am familiar with that workflow. I have used it for a while, tbh I didn't really like it. With interactive rebase I can do the same much more easily, but that's just personal preference. Also, these things don't necessarily need to be done during the PR, only in the end! Please don't spend too much time rewriting the history during development, only here it was necessary, I guess (since there were no rebases to the main branch).

Thank you for adjusting the commit for **Grp**. I have now also added the (final?) commit where the LaTeX macros are added in. I didn't check every formula, but like 20 examples, and they looked fine. 

So when you are ready, I will smash that merge button!

dschepler:
> So when you are ready, I will smash that merge button!

Sounds good to me!

ScriptRaccoon (approved):
✅

ScriptRaccoon:
The proof that **FreeAb** does not have effective cocongruences seems to be generalizable, it looks very formal. Maybe a connection with balanced, here applied to $2 : Z \to Z$. We already have "effective congruences + extensive => mono-regular".

---

We will let $E$ be the abelian group with presentation $\langle a, b, c \mid a - b = 2c \rangle$, with two morphisms $Z \rightrightarrows E$ given by $1\mapsto a$, $1\mapsto b$. Note that $E$ is free with basis $\{ b, c \}$. Then $Hom(E, G) \cong \\{ (x, y, z) \in G^3 \mid x - y = 2z \\}$. Observe that since $G$ is torsion-free, the projection onto the first two coordinates is injective; and $(x, y)$ is in the image precisely when $x \equiv y \pmod{2G}$, which gives an equivalence relation. Therefore, $E$ gives a cocongruence on $Z$.
      On the other hand, if $E$ were the cokernel pair of $h : H \to Z$, that would mean that for $x, y : Z \to G$, $x \equiv y \pmod{2G}$ if and only if $x \circ h = y \circ h$. In particular, from the case $G := Z$, $x := 2 id$, $y := 0$, we would have $2h = 0$. That implies $h = 0$, but then that would give $id_{Z} \equiv 0 \pmod{2}$, resulting in a contradiction.

dschepler:
> The proof that **FreeAb** does not have effective cocongruences seems to be generalizable, it looks very formal. Maybe a connection with balanced, here applied to 2 : Z → Z . We already have "effective congruences + extensive => mono-regular".

Maybe on the dual side, this generalizes to a proof of: additive + effective congruences => normal, forming a revision of the current additive_effective_congruences_imply_normal. The revised proof: given $i : X \hookrightarrow Y$, form the congruence of $\equiv \pmod{i(X)}$ as $E := X \times Y$, with one map being $(x, y) \mapsto y+i(x)$ and the other map being $(x,y) \mapsto y$.

I can do a quick PR for that this evening.

PR #127, opened 2026-04-23
Add the categories of normed and semi-normed vector spaces (WIP)

ScriptRaccoon:
TBA

**TODO.** check if these are locally presentable. This is a blocker since it would imply many other properties which I do not want to assign manually. Asked on MSE: https://math.stackexchange.com/questions/5134166/

Asked on MO: https://mathoverflow.net/questions/511025

**Paused.** I do not understand the answers on MO; maybe someone else can


ScriptRaccoon:
I will close here, since this PR has been inactive for a very long time and it targets a very old version of CatDat.

The mathematical content has more or less been copied to #14, so that this can be reused.

PR #128, opened 2026-04-23
Support display mode and commutative diagrams

ScriptRaccoon:
With this PR we can render formulas in display mode by using the delimiters `$$ ... $$`. The [katex documentation](https://katex.org/docs/options.html) explains what display mode means.

In display mode we can also typeset commutative diagrams via the [amscd package](https://www.jmilne.org/not/Mamscd.pdf).

For example, this commutative diagram

Bildschirmfoto 2026-04-23 um 21 58 16

can be typeset via:

```
$$\begin{CD} X @>{\mathrm{id}}>> X \\ @V{\mathrm{id}}VV @VV{f}V \\ X @>>{f}> Y \end{CD}$$
```

## Diagrams

This PR has replaced all "manual" array diagrams with amscd diagrams. They look much better now.

proof that Delta is cosifted<br />

## Align environments

The `align` environment is also possible in display mode.

isomorphism chain in display mode<br />

## Display mode

This PR also uses display mode in several existing formulas (probably not all) where it was intended anyway. 

typical proof with displaymode formula

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -239,5 +239,5 @@
 		".netlify",
 		"build"
 	],
-	"cSpell.ignoreRegExpList": ["\\$[^$]*\\$"]
+	"cSpell.ignoreRegExpList": ["\\$[^$]*\\$", "\\$\\$[^$]*\\$\\$"]
 }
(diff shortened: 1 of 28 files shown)


varkor:
(If you wanted to render more sophisticated diagrams, you could make use of [quiver](https://github.com/varkor/quiver). I'd be happy to make some changes to make it more convenient to make use of in CatDat.)

ScriptRaccoon:
> (If you wanted to render more sophisticated diagrams, you could make use of [quiver](https://github.com/varkor/quiver). I'd be happy to make some changes to make it more convenient to make use of in CatDat.)

Thanks! I will come back to this as soon as more complicated diagrams pop up. For now, they are quite simple, and amscd is sufficient. That might change in the future.

PR #129, opened 2026-04-24
Database Restructuring

ScriptRaccoon:
After functors have been added to CatDat, several objects in the repository (tables, components, variables, functions, etc.) that only applied to categories previously have been renamed to clarify that categories and functors are on equal grounds. For example, there is a table `category_property_assignments`, and likewise a table `functor_property_assignments`. However, this renaming was never complete. There is still the table `properties` for **category properties**, but `functor_properties` for **functor properties**. This is annoying, confusing, and makes the planned unification of both categorical structures more difficult.

This PR renames the tables as follows:

- `properties` -> `category_properties`
- `related_properties` -> `related_category_properties`
- `implications` -> `category_implications`
- `implication_assumptions` -> `category_implication_assumptions`
- `implication_conclusions` -> `category_implication_conclusions`
- `implications_view` -> `category_implications_view`
- `category_tags` -> `category_tag_assignments`

Also, the auxiliary views `implication_input` and `functor_implication_input` have been removed, since we can use the existing views (to insert implications). Also, indices for unique *lower* IDs (which is required by Netlify routing) have been created for functors (so far they existed only for categories).

I have tried to just add migrations that rename the tables. This worked to some degree, but the views and indices could not be renamed. This is a limitation of SQLite. So it would have been necessary to delete the old views and indices and recreate them with the new names. This would have led to lots of code duplication in the migration files, and even more "garbage". So I have just cleaned it up and created new migration files from scratch. So essentially, I have ignored their purpose ... But at least, now the table definitions are much clearer and have no mistakes anymore.

This also means that after this PR, the CatDat database needs to be cleared (first `catdat-preview`, then `catdat`) and recreated from scratch. Also, collaborators need to delete their local `catdat.db` file and recreate it using the usual command `pnpm db:update`. It is a bit messy, but I am not sure how to avoid that.

This PR also took the opportunity to restructure the data folder. This explains the huge number of changed files. But most of this is just renaming and relocating.

## Updated diagram

catdat-diagram


ScriptRaccoon:
FYI @ykawase5048  @dschepler . After I merge this PR and you pull the changes, you need to delete your local `catdat.db` file (with `pnpm db:kill` for example). Then recreate it using `pnpm db:update`. Ongoing work needs to be rebased, and merge conflicts are likely. Sorry for the inconveniences.

PR #133, opened 2026-04-25
Add the walking splitting

### LINKED ISSUE #123: Add the walking section 🎳
The walking section consists of two objects $0,1$ and morphisms $i : 0 \to 1$, $p : 1 \to 0$ satisfying $pi = \mathrm{id}_0$. The morphism $ip : 1 \to 1$ is idempotent and satisfies $p \circ ip = p$, $ip \circ i = i$. So in total there are 5 morphisms.

diagram of walking section<br />

Let's add this category to the database and decide all of its properties. Related categories in the database are the walking isomorphism and the walking idempotent.

Related: https://github.com/ScriptRaccoon/CatDat/issues/122

ScriptRaccoon:
This PR implements #123 by adding the walking spliting, the category generated by two morphism $i : 0 \to 1$, $p : 1 \to 0$ satisfying $pi = \mathrm{id}_0$.

diagram<br />

I find it quite amusing that this category is pointed and preadditive. It is a skeleton of the category of $\mathbb{F}_2$-vector spaces of dimension $\leq 1$.

All properties have been decided.

### DIFF
--- a/databases/catdat/data/001_categories/008_tiny.sql
+++ b/databases/catdat/data/001_categories/008_tiny.sql
@@ -120,4 +120,14 @@ VALUES
 	'The name of this category comes from the fact that a functor out of it is the same as an <a href="https://ncatlab.org/nlab/show/idempotent" target="_blank">idempotent morphism</a> in the target category. It can also be seen as the delooping of the monoid $\{1,e\}$ in which $e^2=e$.',
 	NULL,
 	NULL
+),
+(
+	'walking_splitting',
+	'walking splitting',
+	'$\mathrm{Split}$',
+	'two objects $0$ and $1$',
+	'the identities, morphisms $i : 0 \to 1$, $p : 1 \to 0$ satisfying $pi = \mathrm{id}_0$, and the idempotent $ip : 1 \to 1$.',
+	'This category could also be called the "walking split idempotent" (or "walking section", "walking retraction"), but we chose a name that emphasizes that the splitting belongs to the data. Notice that the $5$ given morphisms are indeed closed under composition. For example, $p \circ ip = p$ and $ip \circ i = i$.<br>The walking splitting can be interpreted as a skeleton of the category of $\mathbb{F}_2$-vector spaces of dimension $\leq 1$.',
+	NULL,
+	NULL
 );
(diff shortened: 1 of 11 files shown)


ScriptRaccoon:
@ykawase5048  Do you know if this category has sifted colimits?

ykawase5048:
> @ykawase5048 Do you know if this category has sifted colimits?

It seems that this category does not have pullbacks, so it is hard to determine whether it has sifted colimits, though I expect that it does.
Probably, many aspects of sifted colimits are still not well understood, even among experts.

ScriptRaccoon:
@ykawase5048 Yes pullbacks do not exist (we have a terminal object, but no binary products). Can we say if the subcategory $\mathbf{Vect}^{\leq 1} \hookrightarrow \mathbf{Vect}$ is closed under sifted colimits? (Which would be stronger.) Here the supscript means spaces of dimension $\leq 1$. I already checked (this is contained in this PR) that this inclusion preserves filtered colimits.

ScriptRaccoon:
@ykawase5048  I have found a simple proof. Can you check it please?

proof<br />

I have merely used the following property of a sifted category: for finitely many objects $i_1,\dotsc,i_n$ there is an object $k$ with morphisms $i_1 \to k, \dotsc, i_n \to k$. This follows by induction, right? (I am asking because I am rather new to sifted categories and are never sure if I do something which only works for filtered categories.)

PS: it should then also follow that this category is a generalized variety.

ykawase5048:
> @ykawase5048 I have found a simple proof. Can you check it please?
> 
> proof
> 

The proof looks correct and elegant to me!  I didn’t expect that such a linear-algebraic argument would work here.

> I have merely used the following property of a sifted category: for finitely many objects i 1 , … , i n there is an object k with morphisms i 1 → k , … , i n → k . This follows by induction, right? (I am asking because I am rather new to sifted categories and are never sure if I do something which only works for filtered categories.)

Yes, it follows.
 
> PS: it should then also follow that this category is a generalized variety.

We need an additional argument to show that every object in this category is atomic with respect to sifted colimits (strongly finitely presentable), but it seems easy to verify.

PR #145, opened 2026-05-01
Fix formulas causing a scroll instead of breaking

### LINKED ISSUE #135: Fix word breaks in katex formulas
Most formulas look OK:

<table><tr><td>Image</td></tr></table>

But in certain viewport widths, some formulas break, which looks weird.

<table><tr><td>Image</td></tr></table>

When I first saw this problem, I simply added 

```css
.katex {
    white-space: nowrap;
}
```

to the global CSS file (`src/routes/app.css`). It fixes the problem in the example above:

<table><tr><td>Image</td></tr></table>

However, this causes a different problem: in small viewports the formulas can overflow, so that parts of them become invisible.

<table><tr><td>Image</td></tr></table>
 
This is why I have removed the CSS declaration again. Maybe there is a solution, though.

On the other hand, In #128 I already added that CSS line to formulas in display mode since they break more often and this solution is actually mentioned on the katex webpage (https://katex.org/docs/issues). The only other mentioned alternative there is to allow horizontal scrolling, which however did not work (and would be weird in either case).

MoinJulian:
This pull request focuses on improving the display and styling of popup components and mathematical content in the application. The main changes are related to the structure and appearance of the popup's HTML content and better handling of line breaks in rendered math.

**Popup component improvements:**

* Wrapped the popup's dynamic HTML content in a new `<div class="html">` to provide a dedicated styling hook and improve structure.
* Added a CSS rule for `.html` to enable vertical scrolling when the content overflows, ensuring better usability for long popups.

**Math rendering enhancements:**

* Added a `.katex-break` CSS class to allow proper line breaks and inline display for math elements, improving the readability of mathematical content.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
 		"@fortawesome/free-solid-svg-icons": "^7.2.0",
 		"@libsql/client": "^0.17.2",
 		"@octokit/app": "^16.1.2",
+		"dompurify": "^3.4.2",
 		"es6-crawler-detect": "^4.0.2",
 		"ioredis": "^5.10.1",
 		"katex": "^0.16.44",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -24,6 +24,9 @@ importers:
       '@octokit/app':
         specifier: ^16.1.2
         version: 16.1.2
+      dompurify:
+        specifier: ^3.4.2
+        version: 3.4.2
       es6-crawler-detect:
         specifier: ^4.0.2
         version: 4.0.2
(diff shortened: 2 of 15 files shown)


MoinJulian:
This is now basically the same functionality as the current deployed version
https://catdat.app/category/fld

ScriptRaccoon:
The formulas still break. Also, there is a big redundant scrollbar now, also in the wrong direction (`overflow-y: scroll` should be `overflow-x: auto`, if at all).

But scrolling should be the exception; details [here](https://github.com/ScriptRaccoon/CatDat/issues/135#issuecomment-4360573836).

Screenshot from http://localhost:5173/category/walking_splitting:
 
screenshot with breaking formulas<br />

The solution will not be straight forward, otherwise I would have already added the few lines of CSS. ;-)

MoinJulian:
I will have another look at it tomorrow. I tried only breaking it at the = but that doesn't work because it isn't a distinct element and just part of the bin operators.

Just for clarification it should only allow breaking on = but in the best case always stay together in one line?

ScriptRaccoon:
Let's address the XSS in a different PR, i.e. please remove 6ae3514d9b5e704a98676cafcfebff99537e12d6 from this PR. (And please check out my comment at #146 before working on this.)

> Just for clarification it should only allow breaking on = but in the best case always stay together in one line?

Yes. Maybe other relation symbols are also OK, like $\cong$. But tbh I have no idea if this is possible at all. I was only thinking out loud there.

ScriptRaccoon:
Do you plan to continue working on this?

ScriptRaccoon:
I will close this due to inactivity. Let me know if I should reopen.

PR #159, opened 2026-05-05
Fix update script

ScriptRaccoon:
The development server stopped showing data changes after the database was updated with `pnpm db:update`. The cause was #138, and specifically 1772e4b9fcef006fe0d1dd15645a7a493eed6e95 (removal of migrations). In this change, the script was modified to also delete the local database file before rebuilding it from scratch.

I assume this breaks the database connection in the development server, which is why it no longer reacts to changes.

This PR fixes this issue by excluding the setup step from the update command. A downside is that the setup command now has to be run manually by maintainers whenever the schema changes, with no automatic mechanism in place (since migrations have been removed*). For now, I think this is acceptable, as only a small number of people are actively working on CatDat and they can be informed when schema changes occur.

*Maybe I should bring them back.


ScriptRaccoon:
@dschepler In case you also had issues with the development server, they should be fixed with this PR.

PR #160, opened 2026-05-05
Add category of compact Hausdorff spaces

### LINKED ISSUE #156: Add category of compact Hausdorff spaces
I thought I'd like to take a break from the properties side of things, and add a category example of the category **Comp** of compact Hausdorff spaces. Among the properties which are interesting to me:
* It's a pretopos, but it doesn't have a subobject classifier, nor is it cartesian closed.
* In contrast to the example of **Set**_c of a non-topos pretopos, it is both complete and cocomplete.
* It has a cogenerator $[0,1]$ by Urysohn's lemma.
* It is monadic over Set, but (presumably) not finitary algebraic. (At the very least, the canonical forgetful functor is monadic, but it does not preserve sequential colimits.)
* It is extensive and cocomplete but not infinitary extensive. (I think the latter will be automatically deduced using the Girard's theorem based implication.)
* The implications will auto-deduce that it's balanced, thus automatically proving the interesting theorem that a bijective continuous function between compact Hausdorff spaces is a homeomorphism.

dschepler:
Addresses: #156


dschepler:
Hmm, for the coregular property, I have two possible approaches:
Try to adapt the proof from Top.
Otherwise, a Google search reminded me of Gelfand duality, that the opposite category of **CompHaus** is equivalent to the category of commutative unital $C^*$-algebras; and the nLab page on that category says the category of not necessarily unital or commutative $C^*$-algebras is monadic over Set, and therefore regular. Not sure whether there's a way to get from there to regularity of the subcategory.

ScriptRaccoon:
Thanks for the PR!

For the non-existence of NNO, and hence showing that **CompHaus** is not countably distributive, I suggest to use the lemma `nno_distributive_criterion` that I added a few hours ago (to show that **SemiGrp** has no NNO).

If an NNO exists, it has to be $\beta(N)$, and for every compact Hausdorff space $A$, the canonical morphism 

$$\alpha : \beta(A \times N) \to A \times \beta(N)$$

needs to be a split monomorphism. But it is clearly injective and has dense image. So it would actually be an isomorphism. I don't think that this is true when $A$ is not discrete. But I have no proof, yet.

ScriptRaccoon:
I think coregularity should be easy to prove directly. Don't go via C*-algebras here.

dschepler:
The question you raise regarding a consequence of NNO existing also happens to be a special case of whether the category has cartesian filtered colimits, for the special case of the colimit $[n] \to \beta(N)$.

dschepler:
Regarding the property of having cofiltered-limit-stable epimorphisms: It's interesting that the counterexample from **Set** and **Haus** fails in **CompHaus**. In fact, by the intersection theorem, any such example where it's asking about an intersection of a codirected family of nonempty compact subobjects to the constant 1 is doomed to failure. I don't have any ideas on the general case, though.

dschepler:
How about this: if $I = [0, 1]$, then $I^I$ does exist as an exponential in **Top** with the compact-open topology, it's just not compact. So, we have $Hom(\beta(N\times I), I) \simeq Hom(N\times I, I) \simeq Hom(N, I^I) \not\simeq Hom(\beta(N), I^I) \simeq Hom(\beta(N)\times I, I)$, and hopefully by analyzing the chain and a failure example at the one point, we could come up with a proof that $\beta(N\times I) \not\simeq \beta(N)\times I$. I haven't managed to get through the details yet, however.

Maybe constructing the counterexample of a sequence in $I^I$ with no convergent subnet via recursion of multiplying by $x$ would work. Again, not quite getting through the details on that yet.

dschepler:
I think I've got it now: if a natural numbers object $N$ existed, we could iterate the initial conditions $I \to I\times I$, $x \mapsto (x, x)$ and $I\times I \to I \times I$, $(x, y) \mapsto (x, xy)$ to get a continuous function $N \times I \to I \times I$ such that $(n, x) \mapsto (x, x^n)$ for $n \in \mathbb{N}$. The sequence $(n) \in N$ has a convergent subnet $(n_\lambda)_{\lambda \in \Lambda}$, say with limit $y$, and for any $x\in I$, $(n_\lambda, x) \mapsto (x, x^{n_\lambda})$. So taking limits, $(y, x) \mapsto (x, 0)$ if $x \ne 1$ or to $(x, 1)$ if $x=1$. That contradicts that the composition $x \mapsto (y, x) \mapsto (x, \delta_{x,1}) \mapsto \delta_{x,1}$ would have to be continuous.

ScriptRaccoon:
Epimorphisms might actually be stable under cofiltered limits. This is equivalent to: monomorphisms of commutative unital C*-algebras ( = injective \*-homomorphisms) are stable under filtered colimits – which looks correct, even for all C*-algebras.

More generally, exact cofiltered limits are likely. Also, locally copresentable looks promising.

But let's try to find a purely topological proof.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+	'CompHaus',
+	'generator',
+	TRUE,
+	'The one-point space is a generator because it represents the forgetful functor to $\Set$, which is faithful.'
+),
+(
+	'CompHaus',
+	'complete',
+	TRUE,
+	'It is monadic over $\Set$.'
```
ScriptRaccoon: I suggest to replace this with a direct description of limits and the remark that **CompHaus ---> Top** creates them. You need the concrete description of products in the proofs anyway.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+),
+(
+	'CompHaus',
+	'cocomplete',
+	TRUE,
+	'It is monadic over $\Set$.'
```
ScriptRaccoon: I suggest to only use this as an alternative proof and also give a reference that it is monadic.
As the primary proof, I suggest to say that **CompHaus ---> Top** is a reflective subcategory and find a good reference. And then also say a bit how colimits look like. This seems important since you are using the Stone-Cech compactification below.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+),
+(
+-- TODO: rework this when Barr-exact property is added
+	'CompHaus',
+	'regular',
+	TRUE,
+	'It is monadic over $\Set$ and therefore Barr-exact.'
```
ScriptRaccoon: Please add a reference for monadicity.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+),
+(
+-- TODO: rework this when Barr-exact property is added
+	'CompHaus',
+	'effective congruences',
+	TRUE,
+	'It is monadic over $\Set$ and therefore Barr-exact.'
```
ScriptRaccoon: Please add a reference for monadicity.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+),
+(
+	'CompHaus',
+	'cogenerator',
+	TRUE,
+	'The unit interval $[0, 1]$ is a cogenerator: Suppose we have $f, g : X \to Y$ with $f \ne g$. Choose $x\in X$ such that $f(x) \ne g(x)$. Then by Urysohn''s lemma, there is a function $h : Y \to [0, 1]$ such that $h(f(x)) = 0$ and $h(g(x)) = 1$. Therefore, $h\circ f \ne h\circ g$.'
```
ScriptRaccoon:
```suggestion
	'The unit interval $[0, 1]$ is a cogenerator: Suppose we have $f, g : X \rightrightarrows Y$ with $f \ne g$. Choose $x\in X$ such that $f(x) \ne g(x)$. Then by Urysohn''s lemma, there is a continuous function $h : Y \to [0, 1]$ such that $h(f(x)) = 0$ and $h(g(x)) = 1$. Therefore, $h\circ f \ne h\circ g$.'
```

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+	'CompHaus',
+	'extensive',
+	TRUE,
+	'This follows as for $\Top$ or $\Haus$ since finite coproducts in $\CompHaus$ are formed as disjoint union spaces with the disjoint union topology.'
+),
+(
+	'CompHaus',
+	'epi-regular',
+	TRUE,
+	'First, any epimorphism $f : X\to Y$ is surjective: if not, its image would be a closed subset of $Y$. Then by Urysohn''s lemma, there would be a function $g : Y \to [0, 1]$ which is 0 on the image but is not the zero function; but then $g \circ f = 0 \circ f$, giving a contradiction.<br>
```
ScriptRaccoon:
```suggestion
	'First, any epimorphism $f : X\to Y$ is surjective: if not, its image would be a proper subset of $Y$, which is compact and hence closed. Then by Urysohn''s lemma, there would be a non-zero continuous function $g : Y \to [0, 1]$ which is $0$ on the image; but then $g \circ f = 0 \circ f$, giving a contradiction.<br>
```

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+	Therefore, we see that if $g, h : E \rightrightarrows X$ is the kernel pair of $f$, and $U : \CompHaus \to \Set$ is the forgetful functor, then $U(f)$ is the coequalizer of $U(g)$ and $U(h)$. Since $U$ is fully faithful, that implies $f$ is the coequalizer of $g$ and $h$.'
+),
+(
+	'CompHaus',
+	'well-powered',
+	TRUE,
+	'This is clear from the description of monomorphisms as closed embeddings.'
+),
```
ScriptRaccoon: We can remove this because of http://localhost:5173/category-implication/pretopos_balanced

On the second thought, maybe we should keep it because you are using this in the proof of coregularity below.

databases/catdat/data/004_category-implications/002_limits-colimits-behavior-implications.sql
```diff
@@ -260,6 +260,13 @@ VALUES
 	'This holds by definition of a regular category.',
 	FALSE
 ),
+(
+	'regular_well-powered_well-copowered',
+	'["regular", "well-powered"]',
+	'["well-copowered"]',
+	'The regularity condition gives a bijection between quotients of $X$ and effective congruences on $X$, which is a subset of the subobjects of $X\times X$.',
```
ScriptRaccoon:
```suggestion
	'The regularity condition gives a bijection between the collection of quotients of $X$ and the collection of effective congruences on $X$, which is a subcollection of the collection of subobjects of $X\times X$.',
```

Maybe not exactly like this, but I think you see what I want to make more precise here. The "which is" reference in the 2nd sentence is currently not correct, because it would refer to the bijection.

btw you are a native speaker, right? I am not. So take all of this with a grain of salt. I saw that you can write much better than me.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+	'extensive',
+	TRUE,
+	'This follows as for $\Top$ or $\Haus$ since finite coproducts in $\CompHaus$ are formed as disjoint union spaces with the disjoint union topology.'
+),
+(
+	'CompHaus',
+	'epi-regular',
+	TRUE,
+	'First, any epimorphism $f : X\to Y$ is surjective: if not, its image would be a closed subset of $Y$. Then by Urysohn''s lemma, there would be a function $g : Y \to [0, 1]$ which is 0 on the image but is not the zero function; but then $g \circ f = 0 \circ f$, giving a contradiction.<br>
+	Therefore, we see that if $g, h : E \rightrightarrows X$ is the kernel pair of $f$, and $U : \CompHaus \to \Set$ is the forgetful functor, then $U(f)$ is the coequalizer of $U(g)$ and $U(h)$. Since $U$ is fully faithful, that implies $f$ is the coequalizer of $g$ and $h$.'
```
ScriptRaccoon: $U$ is not fully faithful, maybe you want to use the forgetful functor **Top**? But then you actually you need to prove your claim below that epis are quotient maps.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+	'CompHaus',
+	'semi-strongly connected',
+	TRUE,
+	'Every non-empty compact Hausdorff space is weakly terminal (by using constant maps).'
+),
+(
+	'CompHaus',
+	'coregular',
+	TRUE,
+	'Since $\CompHaus$ is mono-regular and epi-regular, the regular-epi-mono factorization system from the regularity of $\CompHaus$ also serves as a regular-epi-mono factorization system for $\CompHaus^\op$. Therefore, it suffices to show that pushouts preserve (regular) monomorphisms in $\CompHaus$. Thus, suppose we have a pushout square
```
ScriptRaccoon:
```suggestion
	'It suffices to show that pushouts preserve (regular) monomorphisms in $\CompHaus$. Thus, suppose we have a pushout square
```

... since coregular is not defined via factorizations in CatDat. If sufficient (co-)limits exist, it is always just the pushout-stability of regular monos. http://localhost:5173/category-property/coregular
See also other proofs of this property, for example for Top_\*.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+	$$\begin{CD}
+	A @> i >> B \\
+	@V f VV @VV g V \\
+	C @>> j > D,
+	\end{CD}$$
+	with $i : A \hookrightarrow B$ a monomorphism. Then for any pair of distinct elements $c, c'' \in C$, by Urysohn''s lemma there exists $\gamma : C \to [0, 1]$ with $\gamma(c) = 0$ and $\gamma(c'') = 1$. Also, by Tietze''s extension theorem, there exists $\beta : B \to [0, 1]$ such that $\beta \circ i = \gamma \circ f$. By the pushout property, there is a unique $\delta : D \to [0, 1]$ such that $\delta \circ g = \beta$ and $\delta \circ j = \gamma$. Since $\delta(j(c)) \ne \delta(j(c''))$, we conclude that $j(c) \ne j(c'')$. This shows that $j$ is injective, so it is a regular monomorphism.'
```
ScriptRaccoon: Nice proof! I suppose we can make an implication about it.

> If a category has an injective cogenerator, then monomorphisms are stable under pushouts.

But let's not add it for now. I will open an issue so that we don't forget, though.

https://github.com/ScriptRaccoon/CatDat/issues/162

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+),
+(
+	'CompHaus',
+	'Malcev',
+	FALSE,
+	'This is clear since $\FinSet$ is not Malcev and can be interpreted as the subcategory of discrete spaces.'
```
ScriptRaccoon:
```suggestion
	'This is clear since $\FinSet$ is not Malcev and can be interpreted as the subcategory of finite discrete spaces.'
```

(I know this is implied by "subcategory", but let's add this for the sake of clarity.)

databases/catdat/data/006_special-morphisms/003_monomorphisms.sql
```diff
@@ -80,6 +80,11 @@ VALUES
 	'injective homomorphisms',
 	'This holds in every finitary algebraic category: the forgetful functor to $\Set$ is faithful, hence reflects monomorphisms, and it is continuous (even representable), hence preserves monomorphisms.'
 ),
+(
+	'CompHaus',
+	'injective continuous maps (which are automatically closed embeddings)',
```
ScriptRaccoon: I know it's "just topology", but can you add the proof that they are closed embeddings?

databases/catdat/data/006_special-morphisms/004_epimorphisms.sql
```diff
@@ -81,6 +81,11 @@ VALUES
 	'a homomorphism of algebras which is an epimorphism of commutative rings',
 	'The forgetful functor $\CAlg(R) \to \Ring$ is faithful and hence reflects epimorphisms, but it also preserves epimorphisms since it preserves pushouts (since $\CAlg(R) \cong R / \Ring$). For epimorphisms of commutative rings see their <a href="/category/CRing">detail page</a>.'
 ),
+(
+	'CompHaus',
+	'surjective continuous maps (which are automatically quotient maps)',
```
ScriptRaccoon: I know it's "just topology", but can you add the proof that they are quotient maps?

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+	'CompHaus',
+	'regular subobject classifier',
+	FALSE,
+	'The proof is almost identical to the one for <a href="/category/Haus">$\Haus$</a>.'
+),
+(
+	'CompHaus',
+	'natural numbers object',
+	FALSE,
+	'Let $I := [0, 1]$. If a natural numbers object $N$ existed, then we could iterate the initial conditions $I\to I\times I$, $x \mapsto (x, x)$ and $I\times I \to I \times I$, $(x, y) \mapsto (x, xy)$ to get a continuous function $N \times I \to I \times I$ such that $(n, x) \mapsto (x, x^n)$ for $x\in I$, $n \in \IN$. The sequence $(n) \in N$ has a convergent subnet $(n_\lambda)_{\lambda \in \Lambda}$, say with limit $y$. Thus, for any $x\in I$ and $\lambda \in \Lambda$, we have $(n_\lambda, x) \mapsto (x, x^{n_\lambda})$. Taking limits, we see $(y, x) \mapsto (x, 0)$ if $x \ne 1$ or $(y, x) \mapsto (x, 1)$ if $x = 1$. In other words, $(y, x) \mapsto (x, \delta_{x, 1})$ for all $x\in I$. However, that contradicts the fact that the composition $I \overset{y \times \id}\longrightarrow N\times I \to I\times I \overset{p_2}\longrightarrow I$, $x \mapsto (y, x) \mapsto (x, \delta_{x,1}) \mapsto \delta_{x,1}$, would have to be continuous.'
```
ScriptRaccoon: I have many questions about this proof.

1. You write

> `such that $(n, x) \mapsto (x, x^n)$ for $x\in I$, $n \in \IN$`

The function is defined on `N \times X`. But you write `n \in \mathbb{N}`. So do you use `\mathbb{N} \subseteq N`? Why can we assume that? Are you using the lemma that I mentioned?

Please add more details here.

2. What does "the sequence $(n) \in N$" mean? How is it defined?

3. Can you please use `$$ ... $$` for the long formula in the end?

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+),
+(
+	'CompHaus',
+	'accessible',
+	FALSE,
+	'For any small regular cardinal $\kappa$, consider the $\kappa$-directed system $\kappa \to \CompHaus$, where for ordinal $\alpha < \kappa$ we have $\alpha \mapsto [0, \alpha]$, and for $\alpha \le \beta < \kappa$ the morphism $\alpha \to \beta$ maps to the inclusion map $[0, \alpha] \hookrightarrow [0, \beta]$. Then the direct colimit of the underlying sets is $\kappa$, so the direct colimit in $\CompHaus$ is the Stone-Čech compatification $\tilde\kappa$ of $\kappa$.<br>
```
ScriptRaccoon:
```suggestion
	'For any small regular cardinal $\kappa$, consider the $\kappa$-directed system $\kappa \to \CompHaus$, where for ordinal $\alpha < \kappa$ we have $\alpha \mapsto [0, \alpha]$, and for $\alpha \le \beta < \kappa$ the morphism $\alpha \to \beta$ maps to the inclusion map $[0, \alpha] \hookrightarrow [0, \beta]$. Then the direct colimit of the underlying spaces is $\kappa$, so the direct colimit in $\CompHaus$ is the Stone-Čech compatification $\tilde\kappa$ of $\kappa$.<br>
```

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,125 @@
...
+	We now claim that any nonempty object $K$ of $\CompHaus$ is not $\kappa$-presentable for any small regular cardinal $\kappa$. If so, then choose a point $x \in \tilde \kappa \setminus \kappa$, and consider the constant map $K \to \tilde \kappa$ with value $x$. By the assumption, this would have to factor through $[0, \alpha]$ for some ordinal $\alpha < \kappa$. But since the topological space $\kappa$ is Tychonoff, the map $\kappa \to \tilde \kappa$ is injective; thus, the map $[0, \alpha] \to \tilde\kappa$ is also injective. That means that the factor must be a constant map with value in $[0, \alpha]$, giving a contradiction.<br>
```
ScriptRaccoon:
```suggestion
	We now claim that any nonempty object $K$ of $\CompHaus$ is not $\kappa$-presentable for any small regular cardinal $\kappa$. If so, then choose a point $x \in \tilde \kappa \setminus \kappa$, and consider the constant map $K \to \tilde \kappa$ with value $x$. By the assumption, this would have to factor through $[0, \alpha]$ for some ordinal $\alpha < \kappa$. But since the topological space $\kappa$ is Tychonoff, the map $\kappa \to \tilde \kappa$ is injective; thus, the map $[0, \alpha] \to \tilde\kappa$ is also injective. That means that the factor must be a constant map with value $x$ in $[0, \alpha]$, giving a contradiction.<br>
```

not sure if you meant this?

databases/catdat/data/003_category-property-assignments/Haus.sql
```diff
@@ -81,7 +81,7 @@ VALUES
...
+	'The proof is similar to <a href="/category/Met">$\Met$</a>. For $n \geq 1$ let $X_n$ be the pushout of $(-\infty, -1/n] \cup [1/n, \infty) \hookrightarrow \IR$ with itself. That is, $X_n$ is the union of two lines $\IR \times \{1\}$ and $\IR \times \{2\}$ where we identify $(x,1) \equiv (x,2)$ when $|x| \geq 1/n$. Then $X_n$ is Hausdorff, and there is a canonical surjective continuous map $X_n \to X_{n+1}$. The colimit in $\Top$ is the union of two lines where we identify $(x,1) \equiv (x,2)$ when $|x| \geq 1/n$ for some $n$, i.e. when $x \neq 0$. This is the line with the double origin, which is not Hausdorff. Its Hausdorff reflection is the line $\IR$ where all points of both lines are identified, and it provides the colimit in $\Haus$. Now, the injective continuous maps $\{1,2\} \to X_n$, $i \mapsto (0,i)$ (where $\{1,2\}$ is discrete) become the constant map $0 : \{1,2\} \to \IR$ in the colimit, which is not a monomorphism.'
```
ScriptRaccoon: fix overflow:

```suggestion
	'The proof is similar to <a href="/category/Met">$\Met$</a>. For $n \geq 1$ let $X_n$ be the pushout of
	$$(-\infty, -1/n] \cup [1/n, \infty) \hookrightarrow \IR$$
	with itself. That is, $X_n$ is the union of two lines $\IR \times \{1\}$ and $\IR \times \{2\}$ where we identify $(x,1) \equiv (x,2)$ when $|x| \geq 1/n$. Then $X_n$ is Hausdorff, and there is a canonical surjective continuous map $X_n \to X_{n+1}$. The colimit in $\Top$ is the union of two lines where we identify $(x,1) \equiv (x,2)$ when $|x| \geq 1/n$ for some $n$, i.e. when $x \neq 0$. This is the line with the double origin, which is not Hausdorff. Its Hausdorff reflection is the line $\IR$ where all points of both lines are identified, and it provides the colimit in $\Haus$. Now, the injective continuous maps $\{1,2\} \to X_n$, $i \mapsto (0,i)$ (where $\{1,2\}$ is discrete) become the constant map $0 : \{1,2\} \to \IR$ in the colimit, which is not a monomorphism.'
```

databases/catdat/data/003_category-property-assignments/CompHaus.sql
dschepler: I don't see how we can remove this. The pretopos_balanced inference is used to infer the category is mono-regular; but if I try commenting this out, the system can't infer it's well-powered.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
dschepler: For any natural numbers object $(N, z : 1 \to N, s : N \to N)$ we have the sequence of global sections $s^n \circ z$, which in a concrete category induces a sequence of elements of $U(N)$.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
ScriptRaccoon: it seems you already removed this now. But as mentioned, maybe we should not do it.

dschepler:
I've found a proof of cofiltered-limit-stable epis. A brief outline: First, a lemma that any cofiltered limit of nonempty compact Hausdorff spaces is nonempty. To see this, consider the product, and for $f : i \to j$ in the cofiltered category, let $F_f := \\{ x \in \prod X_i \mid X_f(x_i) = x_j \\}$. Each one is closed, and it's easy to see the family has the finite intersection property (in writing it up, I'd of course expand this point a bit more). Therefore, the intersection of all $F_f$ is nonempty; but that intersection is precisely the limit.

For the main statement: just apply the lemma to the cofiltered limit of inverse images of $\{ y_i \}$.

I'm still not sure whether this is just a special case of a more general argument in disguise, where the more general argument establishes exact cofiltered limits.

dschepler:
I think I'm getting closer to a proof that it's $\aleph_1$-coaccessible, but I'm still fuzzy on the details.

The basic idea: first prove that $[0,1]$ is $\aleph_1$-copresentable. For that, if we have $\lim X_i \to [0,1]$, then for each point of $\lim X_i$, use continuity and filtering to find an open subset of some $X_i$ whose image is within an interval of diameter at most $1/n$. Choosing a finite subcover of the limit, we can assume that all $i$ are the same. Then if we do this for each $n$, then use filtering, that should find an $i$ such that the map factors through $X_i$.

It should then follow that all countable powers of $[0,1]$ are $\aleph_1$-copresentable. And for any compact $X$, it might then be a limit of all possible maps to such countable powers?

ScriptRaccoon:
I found an interesting paper: https://arxiv.org/pdf/1808.09738
I haven't read it. But it proves an interesting characterization of the category of compact Hausdorff spaces. Also, it gives references for the more well-known properties of this category. Maybe we can exchange the nLab references with them. And maybe this paper also contains proofs for the properties that are currently open.

ScriptRaccoon:
https://doi.org/10.1016/j.topol.2019.02.033 proves several results about locally copresentable categories of spaces (called: dually locally presentable categories). See Theorem 3.4 and Theorem 4.9. It appears that **CompHaus** is never mentioned explicitly, but I assume this is because the authors are way past that example :D. I will find a better reference.

Close catch: https://arxiv.org/pdf/1508.07750 - references in particular the result that **CompHaus**<sup>op</sup> is monadic over **Set**. So I assume the question is if this monad is accessible.

ScriptRaccoon:
Apparently, Isbell proved that **CompHaus**<sup>op</sup> is equivalent to the category of functors $T \to Set$ preserving countable products, where $T$ is the ess. small subcategory of **CompHaus** consisting of all countable powers of the unit interval. So **CompHaus**<sup>op</sup> is actually countable-ary algebraic. From here, it should be clear that the category is locally $\aleph_1$-presentable.

J. Isbell. Generating the algebraic theory of C(X). Algebra Universalis, 15(2):153–155, 1982

I don't have access to the paper. And tbh the papers of Isbell are never easy to understand. So I would appreciate if we can find a more direct argument for local $\aleph_1$-presentability of **CompHaus**<sup>op</sup>.

ScriptRaccoon:
The introduction of http://www.tac.mta.ca/tac/volumes/33/12/33-12.pdf gives useful references.

- in [Dus69] it is proved that the representable functor hom(−, [0, 1]): CompHaus<sup>op</sup> →
Set is monadic,
- the unit interval [0, 1] is shown to be a ℵ1-copresentable compact Hausdorff space
in [GU71],
- a presentation of the algebra operations of CompHaus<sup>op</sup>  is given in [Isb82], and
- a complete description of the algebraic theory of CompHaus<sup>op</sup>  is obtained in [MR17].

dschepler:
To prove a category has exact cofiltered limits, is it sufficient to show that cofiltered limits commute with binary coproducts, initial objects, and coequalizers? Or is there something subtle there that I'm missing?

dschepler:
I think I have a proof that the functor Hom(-, [0,1]) : CompHaus^op -> Set is monadic, using the crude monadicity criterion: suppose we have a coreflexive equalizer $E \to A \rightrightarrows B$ with $r : B \to A$. My mental picture of this uses $r$ to view $B$ as a bundle of compact spaces over $A$, and $f, g : A \rightrightarrows B$ as two sections of that bundle. Then in $Hom(B,[0,1]) \rightrightarrows Hom(A,[0,1]) \to Hom(E,[0,1])$, define $s : Hom(E,[0,1])\to Hom(A,[0,1])$ as a choice function of Tietze extensions. Now given $h : A \to [0,1]$, we can define a continuous function on $im(f)\cup im(g) \simeq A +_E A$ to be $h$ on the first $A$, and to be $s(h|_E)$ on the second $A$. Then again choosing a Tietze extension of this to $B$ for each $h$ gives a function $t : Hom(B,[0,1]) \to Hom(A,[0,1])$. By construction, $s,t$ make $Hom(E,[0,1])$ a split coequalizer.

As for showing it is conservative: suppose $f : X \to Y$ such that $Hom(Y,[0,1]) \to Hom(X,[0,1])$ is a bijection. Then for $x_1\ne x_2$ in $X$, choose a function $\varphi : X \to [0,1]$ with $\varphi(x_1) = 0$, $\varphi(x_2) = 1$. Then since there is $\psi : Y \to [0,1]$ such that $\psi \circ f = \varphi$, we get $f(x_1) \ne f(x_2)$. Therefore, $f$ is injective. Similarly to the proof of epimorphisms being surjective, we can also see $f$ is surjective. So, since $f$ is a bijective continuous map between compact Hausdorff spaces, it is a homeomorphism.

Oh, and for having a left adjoint: that's the functor $S \mapsto [0,1]^S$.

Combined with $\aleph_1$-accessibility of $[0,1]$, that should prove CompHaus^op is locally $\aleph_1$-presentable. (Though I just realized I was missing half of the proof, that if you have two morphisms $X_i \to Y$ which become equal in the limit, then they're already equalized by some $X_f, X_g : X_j \to X_i$. That should be doable by considering sets where the two functions differ by at least $1/n$; use the previous lemma on filtered colimits of nonempty spaces, in contrapositive, to show every point there is not in the image of some $X_f$; use compactness to show the whole part where they differ by at least $1/n$ is not in the image of some $X_f$; then use countable filtering to conclude there's some $X_f$ whose image is entirely in the equalizer.)

ScriptRaccoon:
> To prove a category has exact cofiltered limits, is it sufficient to show that cofiltered limits commute with binary coproducts, initial objects, and coequalizers? Or is there something subtle there that I'm missing?

Yes this works as soon as cofiltered limits and finite colimits exist, otherwise they cannot be computed pointwise e.g. and this is also why it was an assumption for the property in the first place. Also, initial objects are never a problem, as the constant diagram with value X on a non-empty category has limit X.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,140 @@
...
+	'CompHaus',
+	'equalizers',
+	TRUE,
+	'The equalizer in $\Top$ of two continuous functions $f, g : X \rightrightarrows Y$ between compact Hausdorff spaces is a closed subspace of $X$, and therefore it is also compact Hausdorff. Since the forgetful functor from $\CompHaus$ to $\Top$ is fully faithful, this limit is reflected in $\CompHaus$ as well.'
+),
+(
+	'CompHaus',
+	'cocomplete',
+	TRUE,
+	'$\CompHaus$ is a reflective subcategory of $\Top$, with the reflector being the Stone-Čech compatification functor. See <a href="https://ncatlab.org/nlab/show/compact+Hausdorff+space#StoneCechCompactification" target="_blank">nLab</a> for example. Therefore, as usual, we can form colimits in $\CompHaus$ by forming colimits in $\Top$ and then applying Stone-Čech compatification.'
```
ScriptRaccoon: typos: `compatification`

ScriptRaccoon:
I would like to hear your feedback regarding the current state of #164 (the biggest PR I have worked on so far). For this I have converted the definition of **CompHaus** and a selection of its properties from this PR into a yaml file `CompHaus.yaml`, which you see below. I would very much appreciate to know what you think about this format. Is it better than the current model? What is your first impression?

```yaml
id: CompHaus
name: category of compact Hausdorff spaces
notation: $\CompHaus$
objects: compact Hausdorff spaces
morphisms: continuous functions
description: |
  This is the full subcategory of $\Top$ consisting of those spaces that are [compact](https://en.wikipedia.org/wiki/Compact_space) and [Hausdorff](https://en.wikipedia.org/wiki/Hausdorff_space).
nlab_link: https://ncatlab.org/nlab/show/empty+category
tags:
  - topology

related_categories:
  - Top
  - Haus

satisfied_properties:
  - property_id: locally small
    reason: This is trivial.

  - property_id: generator
    reason: The one-point space is a generator because it represents the forgetful functor to $\Set$, which is faithful.

  - property_id: products
    reason: |
      By the Tychonoff product theorem, a product in $\Top$ of compact Hausdorff spaces is compact; it is also clearly Hausdorff.
      Since the forgetful functor from $\CompHaus$ to $\Top$ is fully faithful, this limit is reflected in $\CompHaus$ as well.

  - property_id: equalizers
    reason: |
      The equalizer in $\Top$ of two continuous functions $f, g : X \rightrightarrows Y$ between compact Hausdorff spaces is a closed subspace of $X$, and therefore it is also compact Hausdorff. Since the forgetful functor from $\CompHaus$ to $\Top$ is fully faithful, this limit is reflected in $\CompHaus$ as well.

  - property_id: cocomplete
    reason: |
      $\CompHaus$ is a reflective subcategory of $\Top$, with the reflector being the Stone-Čech compactification functor.
      See [nLab](https://ncatlab.org/nlab/show/compact+Hausdorff+space#StoneCechCompactification) for example.
      Therefore, as usual, we can form colimits in $\CompHaus$ by forming colimits in $\Top$ and then applying Stone-Čech compactification.

unsatisfied_properties:
  - property_id: natural numbers object
    reason: |
      Let $I := [0, 1]$. If a natural numbers object $(N, z : 1 \to N, s : N \to N)$ existed, then we could iterate the initial conditions $I\to I\times I$, $x \mapsto (x, x)$ and the recursive step function $I\times I \to I \times I$, $(x, y) \mapsto (x, xy)$ to get a continuous function $N \times I \to I \times I$ such that $(s^n(z), x) \mapsto (x, x^n)$ for $x\in I$, $n \in \IN$.
      The sequence $(s^n(z)) \in N$ has a convergent subnet $(s^{n_\lambda}(z))_{\lambda \in \Lambda}$, say with limit $y$. Thus, for any $x\in I$ and $\lambda \in \Lambda$, we have $(s^{n_\lambda}(z), x) \mapsto (x, x^{n_\lambda})$.
      Taking limits, we see $(y, x) \mapsto (x, 0)$ if $x \ne 1$ or $(y, x) \mapsto (x, 1)$ if $x = 1$. In other words, $(y, x) \mapsto (x, \delta_{x, 1})$ for all $x\in I$. However, that contradicts the fact that the composition
      $$I \overset{y \times \id}\longrightarrow N\times I \to I\times I \overset{p_2}\longrightarrow I \\ x \mapsto (y, x) \mapsto (x, \delta_{x,1}) \mapsto \delta_{x,1},$$
      would have to be continuous.

special_objects:
  terminal object:
    description: singleton space
  initial object:
    description: empty space

special_morphisms:
  epimorphisms:
    description: surjective continuous maps (which are automatically quotient maps)
    reason: For the non-trivial direction, and for a proof of the parenthetical remark, see the proof above that $\CompHaus$ is epi-regular.
```

dschepler:
> I would like to hear your feedback regarding the current state of #164 (the biggest PR I have worked on so far). For this I have converted the definition of **CompHaus** and a selection of its properties from this PR into a yaml file `CompHaus.yaml`, which you see below. I would very much appreciate to know what you think about this format. Is it better than the current model? What is your first impression?

The sample looks great, and I agree with the comments by others that having everything directly related to the category in one file should be easier to work with.

ScriptRaccoon:
Thank you for the feedback!

The mentioned PR is a *lot* of work and I wanted to make sure that I am going into the right direction. Apparently, yes.

ScriptRaccoon:
... but I have decided against using markdown (long story...), so instead of 
`[Hausdorff](https://en.wikipedia.org/wiki/Hausdorff_space)`
it will be
`<a href="https://en.wikipedia.org/wiki/Hausdorff_space" target="_blank">Hausdorff</a>`
as before. So the end result will be a bit different from the file I shared above.

dschepler:
Interesting search...
image
(And the dual search returns no results.)

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,66 @@
...
+\author{Daniel Schepler}
+\date{\today}
+
+\begin{document}
+\maketitle
+
+We let $\CompHaus$ denote the category of compact Hausdorff spaces.
+
+\begin{lemma}
+	The functor $\Hom({-}, [0, 1])) : \CompHaus^{\op} \to \Set$ is monadic.
```
ScriptRaccoon:
```suggestion
	The functor $\Hom({-}, [0, 1]) : \CompHaus^{\op} \to \Set$ is monadic.
```

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,66 @@
...
+	
+	\begin{proof}
+		We use the crude monadicity theorem. First, the functor has a left adjoint $S \mapsto [0, 1]^S$ with the evident isomorphism
+		$$\Hom_{\CompHaus}(X, [0, 1]^S) \simeq \Hom_{\Set}(S, \Hom_{\CompHaus}(X, [0, 1])). $$
+		
+		To see the functor is conservative, suppose we have a continuous function $f : X \to Y$ such that $f^* : \Hom(Y, [0, 1]) \to \Hom(X, [0, 1])$ is a bijection. Then for any $x_1, x_2 \in X$ with $x_1 \ne x_2$, there exists $\varphi : X \to [0, 1]$ with $\varphi(x_1) = 0$ and $\varphi(x_2) = 1$. Since $f^*$ is surjective, there exists $\psi : Y \to [0, 1]$ with $\varphi = \psi \circ f$; thus, we must have $f(x_1) \ne f(x_2)$. Likewise, we know the image of $f$ is closed. If this image is not all of $Y$, then there exists nonzero $\varphi : Y \to [0, 1]$ which is zero on the image. But then $\varphi \circ f = 0 \circ f$, contradicting the injectivity of $f^*$. Thus, $f$ is a bijective continuous function, and therefore a homeomorphism.
```
ScriptRaccoon: please cite Urysohn's Lemma and Tietze's theorem where appropriate

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,66 @@
...
+\begin{document}
+\maketitle
+
+We let $\CompHaus$ denote the category of compact Hausdorff spaces.
+
+\begin{lemma}
+	The functor $\Hom({-}, [0, 1])) : \CompHaus^{\op} \to \Set$ is monadic.
+	
+	\begin{proof}
+		We use the crude monadicity theorem. First, the functor has a left adjoint $S \mapsto [0, 1]^S$ with the evident isomorphism
```
ScriptRaccoon: Let's add a reference that proves the crude monadicity theorem. For some reason, I could not find one. In Barr-Wells' TTT it is just said that it is clear. :D

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,146 @@
...
+	'CompHaus',
+	'cofiltered-limit-stable epimorphisms',
+	TRUE,
+	'Suppose we have a cofiltered diagram of epimorphisms $(f_i : X_i \to Y_i)$, and $y = (y_i) \in \lim_i Y_i$. Then by <a href="/lemma/cofiltered-limit-of-non-empty-compact">this result</a>, the limit of $f_i^{-1}(\{ y_i \})$ is non-empty. If $x$ is in this limit, that implies that $(\lim_i f_i)(x) = y$.'
+),
+(
+	'CompHaus',
+	'locally copresentable',
+	TRUE,
+	'A proof can be found <a href="/pdf/comphaus_copresentable.pdf">here</a>.'
```
ScriptRaccoon: Please also add an original reference for this result (see my comments). Either here or in the pdf. Otherwise, readers might get a wrong impression with regards to orginality.

In any case, I think it is *very good* if we offer a well-written proof directly in CatDat. Thank you for adding it!

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,66 @@
...
+		$$\Hom_{\CompHaus}(X, [0, 1]^S) \simeq \Hom_{\Set}(S, \Hom_{\CompHaus}(X, [0, 1])). $$
+		
+		To see the functor is conservative, suppose we have a continuous function $f : X \to Y$ such that $f^* : \Hom(Y, [0, 1]) \to \Hom(X, [0, 1])$ is a bijection. Then for any $x_1, x_2 \in X$ with $x_1 \ne x_2$, there exists $\varphi : X \to [0, 1]$ with $\varphi(x_1) = 0$ and $\varphi(x_2) = 1$. Since $f^*$ is surjective, there exists $\psi : Y \to [0, 1]$ with $\varphi = \psi \circ f$; thus, we must have $f(x_1) \ne f(x_2)$. Likewise, we know the image of $f$ is closed. If this image is not all of $Y$, then there exists nonzero $\varphi : Y \to [0, 1]$ which is zero on the image. But then $\varphi \circ f = 0 \circ f$, contradicting the injectivity of $f^*$. Thus, $f$ is a bijective continuous function, and therefore a homeomorphism.
+		
+		Finally, suppose we have a coreflexive equalizer pair $E \xhookrightarrow{i} A \overset{f}{\underset{g}{\rightrightarrows}} B$ with $r : B \to A$. We may assume that $i$ is a subspace inclusion map. We may use $r$ to think of $B$ as a bundle of compact spaces over $A$, with two sections $f, g$. We then need to show that
```
ScriptRaccoon:
```suggestion
		Finally, suppose we have a coreflexive equalizer pair
		$$E \xhookrightarrow{i} A \overset{f}{\underset{g}{\rightrightarrows}} B$$
		with $r : B \to A$. We may assume that $i$ is a subspace inclusion map. We may use $r$ to think of $B$ as a bundle of compact spaces over $A$, with two sections $f, g$. We then need to show that
```

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,66 @@
...
+	\end{proof}
+\end{lemma}
+
+This shows that $\CompHaus^{\op}$ is equivalent to the category of algebras over the monad $S \mapsto \Hom_{\CompHaus}([0, 1]^S, [0, 1])$. We may view such algebras as being models of the algebraic theory of all continuous functions $[0,1]^S \to [0,1]$. In fact, we can show that any such function only depends on countably many coordinates in the domain, so that operations of this theory will be generated by the continuous functions $[0,1]^\omega \to [0,1]$.
```
ScriptRaccoon:
```suggestion
This shows that $\CompHaus^{\op}$ is equivalent to the category of algebras over the monad $S \mapsto \Hom_{\CompHaus}([0, 1]^S, [0, 1])$. We may view such algebras as being models of the infinitary algebraic theory of all continuous functions $[0,1]^S \to [0,1]$. In fact, we can show that any such function only depends on countably many coordinates in the domain, so that operations of this theory will be generated by the continuous functions $[0,1]^\omega \to [0,1]$.
```

ScriptRaccoon: > In fact, we can show that any such function only depends on countably many coordinates in the domain, so that operations of this theory will be generated by the continuous functions $[0,1]^\omega \to [0,1]$.

It is not clear for the reader at this point if you are going to prove this and if we need this result to reach the goal.

It is a consequence of the Lemma, right? So let us maybe add a short sentence making this connection.

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,66 @@
...
+
+\begin{lemma}
+	The object $[0,1]$ of $\CompHaus$ is $\aleph_1$-copresentable.
+	
+	\begin{proof}
+		Suppose we have an $\aleph_1$-cofiltered limit $X = \lim_{i\in \I} X_i$ with projections $p_i : X \to X_i$, and a continuous function $\varphi : X \to [0,1]$. Then for any $x\in X$ and $n\in \INnz$, there exists an interval neighborhood $N_x$ of $\varphi(x)$ of diameter at most $1/n$, along with a basic open neighborhood $\bigcap_{a=1}^b p_{i_a}^{-1}(U_a)$ of $x$ whose image is contained in $N_x$. Taking a cone $(j, f_a : j \to i_a)$ of $i_1, \ldots, i_b$, we may form $V := \bigcap_{a=1}^b X_{f_a}^{-1}(U_a)$; the basic open neighborhood is then equal to $p_j^{-1}(V)$.
```
ScriptRaccoon: Actually, you choose *any* $N_x$, right? And then we choose a neighborhood in the the preimage. Maybe we can reformulate this. As written, it sounds as if the existence of $N_x$ is something which needs to be proven.

ScriptRaccoon: The dependency on $n$ is not clear at this point. Do you fix $n$? Do you do this construction for all $n$?

From the later paragraphs, I think what you are doing is to fix $n$.

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,66 @@
...
+		
+		By compactness of $X$, we may take finitely many such basic open neighborhoods of the form $p_j^{-1}(V)$ which cover $X$. Again using the assumption that $\I$ is cofiltered, we may assume that $j$ is the same for each neighborhood. In particular, we see that whenever we have $x, y\in X$ with $p_j(x) = p_j(y)$, then $|\varphi(x) - \varphi(y)| < 1/n$.
+		
+		Now choose such a $j_n$ for each $n$, and then use the fact that $\I$ is $\aleph_1$-cofiltered to take a cone $(k, f_n : k \to j_n)$. We then see that whenever we have $x, y\in X$ with $p_k(x) = p_k(y)$, then $\varphi(x) = \varphi(y)$. Thus, $\varphi$ induces a well-defined function on the image of $p_k$. This function is continuous, since by construction for any $n\in \INnz$ and $x \in X$, there is a neighborhood $V$ of $p_k(x)$ such that whenever $p_k(y)\in V$ as well, $|p_k(y) - p_k(x)| < 1/n$. This induced function can then can be extended to a continuous function $\psi : X_k \to [0,1]$ such that $\varphi = \psi \circ p_k$.
```
ScriptRaccoon: Typo:
 
```suggestion
		Now choose such a $j_n$ for each $n$, and then use the fact that $\I$ is $\aleph_1$-cofiltered to take a cone $(k, f_n : k \to j_n)$. We then see that whenever we have $x, y\in X$ with $p_k(x) = p_k(y)$, then $\varphi(x) = \varphi(y)$. Thus, $\varphi$ induces a well-defined function on the image of $p_k$. This function is continuous, since by construction for any $n\in \INnz$ and $x \in X$, there is a neighborhood $V$ of $p_k(x)$ such that whenever $p_k(y)\in V$ as well, $|\varphi(x) - \varphi(y)| < 1/n$. This induced function can then can be extended to a continuous function $\psi : X_k \to [0,1]$ such that $\varphi = \psi \circ p_k$.
```

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,66 @@
...
+		
+		Likewise, suppose we have $\varphi, \psi : X_i \to [0,1]$ such that $\varphi \circ p_i = \psi \circ p_i$. For each $n\in \INnz$, consider the set $D_n := \{ x \in X_i \mid |\varphi(x) - \psi(x)| \ge 1/n \}$. For any $x \in D_n$, we must have $x \notin \im(p_i)$. Now recall that we previously proved a result that a cofiltered limit of non-empty compact Hausdorff spaces is nonempty. Using the contrapositive of this, we can conclude that if $x \notin \im(p_i)$, then there exists $f : j \to i$ such that $x \not\in \im(X_f)$. Now $D_n \setminus \im(X_f)$ is open and such sets cover $D_n$; taking a finite subcover and then using the cofiltering assumption again, we conclude that there is a single $f_n : j_n \to i$ such that $\im(X_{f_n})$ is disjoint from $D_n$. Using the $\aleph_1$-cofiltering assumption to take a cone of the $f_n$, we get that there is $f : k \to i$ such that $\im(X_f)$ is disjoint from all $D_n$. This means that $\phi \circ F_f = \psi \circ F_f$.
```
ScriptRaccoon:
```suggestion
		Likewise, suppose we have $\varphi, \psi : X_i \rightrightarrows [0,1]$ such that $\varphi \circ p_i = \psi \circ p_i$. For each $n\in \INnz$, consider the set $D_n := \{ x \in X_i \mid |\varphi(x) - \psi(x)| \ge 1/n \}$. For any $x \in D_n$, we must have $x \notin \im(p_i)$. Now recall that we previously proved a result that a cofiltered limit of non-empty compact Hausdorff spaces is nonempty. Using the contrapositive of this, we can conclude that if $x \notin \im(p_i)$, then there exists $f : j \to i$ such that $x \not\in \im(X_f)$. Now $D_n \setminus \im(X_f)$ is open and such sets cover $D_n$; taking a finite subcover and then using the cofiltering assumption again, we conclude that there is a single $f_n : j_n \to i$ such that $\im(X_{f_n})$ is disjoint from $D_n$. Using the $\aleph_1$-cofiltering assumption to take a cone of the $f_n$, we get that there is $f : k \to i$ such that $\im(X_f)$ is disjoint from all $D_n$. This means that $\phi \circ F_f = \psi \circ F_f$.
```

Also, let's denote the maps differently because $\varphi,\psi$ have completely different meaning than in the preceding paragraph. (Sure, it's not necessary, but I think it will be easier to read.)

ScriptRaccoon: >  Now recall that we previously proved a result that a cofiltered limit of non-empty compact Hausdorff spaces is nonempty.

The pdf needs to be self-contained. Imagine someone gives you link to this pdf. Then this reference is not clear.

Suggestion: Move the lemma from `001_lemmas.sql` (whose proof I reformulated, btw) to a lemma in this pdf. Of course, then in the proof of `cofiltered-limit-stable epimorphisms`, refer to the pdf.

ScriptRaccoon: > Using the contrapositive of this, we can conclude that if $x \notin \im(p_i)$, then there exists $f : j \to i$ such that $x \not\in \im(X_f)$. 

I don't understand this part.

ScriptRaccoon: Last sentence must be:

```
This means that $\phi \circ X_f = \psi \circ X_f$.
```

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,66 @@
...
+	\end{proof}
+\end{lemma}
+
+\begin{corollary}
+	The category $\CompHaus$ is $\aleph_1$-copresentable.
```
ScriptRaccoon:
```suggestion
	The category $\CompHaus$ is locally $\aleph_1$-copresentable.
```

(as in the title)

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,146 @@
...
+	TRUE,
+	'First, any epimorphism $f : X\to Y$ is surjective: if not, its image would be a proper subset of $Y$, which is compact and hence closed. Then by Urysohn''s lemma, there would be a non-zero continuous function $g : Y \to [0, 1]$ which is $0$ on the image; but then $g \circ f = 0 \circ f$, giving a contradiction.<br>
+	Now the identity morphism from $Y$, with the quotient topology of $f$, to $Y$ with its given topology is a bijective continuous function between compact Hausdorff spaces, so it is a homeomorphism. In other words, $f$ is a quotient map. Therefore, we see that if $g, h : E \rightrightarrows X$ is the kernel pair of $f$, and $U : \CompHaus \to \Top$ is the forgetful functor, then $U(f)$ is the coequalizer of $U(g)$ and $U(h)$. Since $U$ is fully faithful, that implies $f$ is the coequalizer of $g$ and $h$.'
+),
+(
+	'CompHaus',
+	'semi-strongly connected',
+	TRUE,
+	'Every non-empty compact Hausdorff space is weakly terminal (by using constant maps).'
+),
```
ScriptRaccoon:
```suggestion
	'CompHaus',
	'semi-strongly connected',
	TRUE,
	'This is already true for <a href="/category/Top">$\mathbf{Top}$</a>.'
),
```
 
maybe?

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,146 @@
...
+	'CompHaus',
+	'regular subobject classifier',
+	FALSE,
+	'The proof is almost identical to the one for <a href="/category/Haus">$\Haus$</a>.'
+),
+(
+	'CompHaus',
+	'natural numbers object',
+	FALSE,
+	'Let $I := [0, 1]$. If a natural numbers object $(N, z : 1 \to N, s : N \to N)$ existed, then we could iterate the initial conditions $I\to I\times I$, $x \mapsto (x, x)$ and the recursive step function $I\times I \to I \times I$, $(x, y) \mapsto (x, xy)$ to get a continuous function $N \times I \to I \times I$ such that $(s^n(z), x) \mapsto (x, x^n)$ for $x\in I$, $n \in \IN$. The sequence $(s^n(z)) \in N$ has a convergent subnet $(s^{n_\lambda}(z))_{\lambda \in \Lambda}$, say with limit $y$. Thus, for any $x\in I$ and $\lambda \in \Lambda$, we have $(s^{n_\lambda}(z), x) \mapsto (x, x^{n_\lambda})$. Taking limits, we see $(y, x) \mapsto (x, 0)$ if $x \ne 1$ or $(y, x) \mapsto (x, 1)$ if $x = 1$. In other words, $(y, x) \mapsto (x, \delta_{x, 1})$ for all $x\in I$. However, that contradicts the fact that the composition
```
ScriptRaccoon: We can just work with a convergent subsequence, right?

databases/catdat/data/003_category-property-assignments/CompHaus.sql
```diff
@@ -0,0 +1,146 @@
...
+),
+(
+	'CompHaus',
+	'exact cofiltered limits',
+	FALSE,
+	'Consider the $\IN$-codirected systems $X_n := [0, 1] \times [0, 1/n]$ with the maps $X_{n+1} \to X_n$ being inclusion maps, and $Y_n := [0, 1+1/n]$ with the maps $Y_{n+1} \to Y_n$ also being inclusion maps. We define $f_n : X_n \to Y_n$, $(x, y) \mapsto x$ and $g_n : X_n \to Y_n$, $(x, y) \mapsto x+y$. It is straightforward to check these give morphisms of $\IN$-codirected systems in $\CompHaus$.<br>
+	Now for each $n$, the coequalizer of $f_n$ and $g_n$ is a single-point space. On the other hand, $\lim X_n \simeq [0, 1] \times \{ 0 \}$; $\lim Y_n \simeq [0, 1]$; and $\lim f_n = \lim g_n$, $(x, 0) \mapsto x$. Thus, the coequalizer of $\lim f_n$ and $\lim g_n$ is $[0, 1]$, showing that this coequalizer is not preserved under limits.'
```
ScriptRaccoon: Can wee add the proof that the coequalizer of $f_n,g_n$ is terminal? I know this is simple analysis, but it is not immediate either.

We need to prove that every (continuous) map $h : Y_n \to T$ with $h(x+y)=h(x)$ for $x \in [0,1]$, $y \in [0,1/n]$ is constant. 

More generally, let $h : [0,a+b] \to T$ satisfy $h(x)=h(x+y)$ for all $x \in [0,a]$, $y \in [0,b]$. We claim that $h$ is constant. We will show that every value is $h(a)$.

Case 1: Values on $[a,a+b]$.  
If $t \in [a,a+b]$, write $t=a+y$ with $y \in [0,b]$. Then $h(t)=h(a+y)=h(a)$.

Case 2: Values on $[0,a]$.  
Let $x \in [0,a]$. Define a finite sequence in $[0,a]$ as follows. Let $x_0 := x$. While $x_n + b \leq a$, let $x_{n+1} := x_n + b$. This yields a finite sequence $x_0,\dotsc,x_k$ in $[0,a]$ with $x_k + b > a$. Then $a - x_k \in [0,b]$, so $h(x_k) = h(x_k + (a - x_k)) = h(a)$. By backwards induction it now follows $h(x_n)=h(a)$, namely: $h(x_{n-1}) = h(x_{n-1} + b) = h(x_n) = h(a)$. Hence, $h(x) = h(a)$.

databases/catdat/data/003_category-property-assignments/CompHaus.sql
dschepler: Definitely not. Any subsequence of $\mathbb{N}$ has infinitely many elements of $\beta\mathbb{N} \setminus \mathbb{N}$ as cluster points (where as you've pointed out, $\beta\mathbb{N}$ would be the obvious candidate for a natural numbers object). The standard construction of a convergent subnet would end up with the subnet indexed by elements of a non-principal ultrafilter of $\mathbb{N}$, ordered by reverse inclusion.

ScriptRaccoon:
I think I am finished with #164.  Should I wait with the merge? As before, I am not sure which order is best. Maybe you have a preference?

dschepler:
> I think I am finished with #164. Should I wait with the merge? As before, I am not sure which order is best. Maybe you have a preference?

It looks like I have a significant amount of revisions to the pdf file (and a few updates otherwise) to make. So, it's probably best if you merge #164 and I can take care of updating my changes to the new YAML schema, rather than my revisions delaying the other merge for maybe a day or two.

ScriptRaccoon:
Alright. I have merged the PR now. If you need any help with transforming the data to a YAML file, or if something is a bit weird, please let me know. The [documentation](https://github.com/ScriptRaccoon/CatDat/blob/main/CONTRIBUTING.md) has been updated as well.

static/pdf/comphaus_copresentable.tex
dschepler: "In fact, we can prove this slightly stronger condition: Lemma..."  Something like that?

I was also thinking of the statement about being generated by finitary and $\aleph_0$-ary functions as closely related to Isbell's result, via applying the general framework of Lawvere theories to this case. In fact, I was already thinking of putting an alternate proof via this route in the last corollary - and if I added it now, it would be a perfect hook to citing that paper.

static/pdf/comphaus_copresentable.tex
dschepler: I was just thinking of $N_x$ as an abstraction of $(\varphi(x) - 1/(2n), \varphi(x) + 1/(2n)) \cap [0, 1]$ -- and to emphasize I only cared about the diameter in the image in $[0,1]$ so if I take a finite subcover in the limit or in $[0,1]$ then it doesn't matter if the interval from the subcover doesn't have a particular point as its center. I'll definitely think about how I could reformulate that.

static/pdf/comphaus_copresentable.tex
dschepler: Maybe it helps to mention I'm using that a slice of a cofiltered category is cofiltered, and the limit over a slice is the same as over the full cofiltered category; and then I'm applying the lemma to show if $f^{-1}(x)$ is non-empty for each $f:j\to i$ then $p_i^{-1}(x)$ is non-empty?

ScriptRaccoon: Yeah the slice argument was clear to me, the rest not. Still not after your comment, tbh. I mean, it will be rather simple, I know, what something is left out here.

static/pdf/comphaus_copresentable.tex
ScriptRaccoon: > "In fact, we can prove this slightly stronger condition: Lemma..." Something like that?

Yes.

I don't have an opinion about the rest since I am not familiar with Isbell's work.

databases/catdat/data/categories/CompHaus.yaml
```diff
@@ -0,0 +1,123 @@
...
+    check_redundancy: false
+
+  - property_id: cocomplete
+    reason: $\CompHaus$ is a reflective subcategory of $\Top$, with the reflector being the Stone-Čech compactification functor. See <a href="https://ncatlab.org/nlab/show/compact+Hausdorff+space#StoneCechCompactification" target="_blank">nLab</a> for example. Therefore, as usual, we can form colimits in $\CompHaus$ by forming colimits in $\Top$ and then applying Stone-Čech compatification.
+
+  - property_id: regular
+    reason: The forgetful functor from $\CompHaus$ to $\Set$ is monadic; see for example <a href="https://ncatlab.org/nlab/show/compact+Hausdorff+space#compact_hausdorff_spaces_are_monadic_over_sets">nLab</a>. Therefore, by <a href="https://ncatlab.org/nlab/show/colimits+in+categories+of+algebras#exact">this result</a>, $\CompHaus$ is Barr-exact and in particular is regular.
+
+  - property_id: effective congruences
```
ScriptRaccoon: Let's add back in the TODO comments for Barr-exact.

databases/catdat/data/categories/CompHaus.yaml
```diff
@@ -0,0 +1,123 @@
...
+    reason: This follows as for $\Top$ or $\Haus$ since finite coproducts in $\CompHaus$ are foemd as disjoint union spaces with the disjoint union topology.
+
+  - property_id: epi-regular
+    reason: |-
+      First, any epimorphism $f : X\to Y$ is surjective: if not, its image would be a proper subset of $Y$, which is compact and hence closed. Then by Urysohn's lemma, there would be a non-zero continuous function $g : Y \to [0, 1]$ which is $0$ on the image; but then $g \circ f = 0 \circ f$, giving a contradiction.
+
+      Now the identity morphism from $Y$, with the quotient topology of $f$, to $Y$ with its given topology is a bijective continuous function between compact Hausdorff spaces, so it is a homeomorphism. In other words, $f$ is a quotient map. Therefore, we see that if $g, h : E \rightrightarrows X$ is the kernel pair of $f$, and $U : \CompHaus \to \Top$ is the forgetful functor, then $U(f)$ is the coequalizer of $U(g)$ and $U(h)$. Since $U$ is fully faithful, that implies $f$ is the coequalizer of $g$ and $h$.
+
+  - property_id: semi-strongly connected
+    reason: This is already true for <a href="/category/Top">$\mathbf{Top}$</a>.
```
ScriptRaccoon: \Top

dschepler:
The spacing on the display of the "natural numbers object" proof is weird now. I could maybe alleviate it a bit by putting them in a single $$ line 1 \\ line 2 $$ display; but then the "pnpm dev" console would constantly be displaying warnings that \\ within $$ is not supported by LaTeX. And that would still result in an extra blank line between the display and the next line.

ScriptRaccoon:
> The spacing on the display of the "natural numbers object" proof is weird now. I could maybe alleviate it a bit by putting them in a single $$ line 1 \ line 2 $$ display; but then the "pnpm dev" console would constantly be displaying warnings that \ within $$ is not supported by LaTeX. And that would still result in an extra blank line between the display and the next line.

How I would do it:

```yaml
  - property_id: natural numbers object
    reason: >-
      Let $I := [0, 1]$. If a natural numbers object $(N, z : 1 \to N, s : N \to N)$ existed, then we could iterate the initial conditions $I\to I\times I$, $x \mapsto (x, x)$ and the recursive step function $I\times I \to I \times I$, $(x, y) \mapsto (x, xy)$ to get a continuous function $N \times I \to I \times I$ such that $(s^n(z), x) \mapsto (x, x^n)$ for $x\in I$, $n \in \IN$. The sequence $(s^n(z)) \in N$ has a convergent subnet $(s^{n_\lambda}(z))_{\lambda \in \Lambda}$, say with limit $y$. Thus, for any $x\in I$ and $\lambda \in \Lambda$, we have $(s^{n_\lambda}(z), x) \mapsto (x, x^{n_\lambda})$. Taking limits, we see $(y, x) \mapsto (x, 0)$ if $x \ne 1$ or $(y, x) \mapsto (x, 1)$ if $x = 1$. In other words, $(y, x) \mapsto (x, \delta_{x, 1})$ for all $x\in I$. However, that contradicts the fact that the composition
      $$\begin{align*} I & \xrightarrow{y \times \id}  N \times I \xrightarrow{} I\times I \xrightarrow{p_2} I, \\ x & \mapsto (y, x) \mapsto (x, \delta_{x,1}) \mapsto \delta_{x,1}, \end{align*}$$
      would have to be continuous.
```
 
This looks fine now.

NNO proof

You can also keep the line break in the alignment. Just use >- as the marker

dschepler:
I've just made more revisions to the PDF to try to clarify the arguments. Let me know if I still need more clarification.

ScriptRaccoon:
The merge conflict comes from https://github.com/ScriptRaccoon/CatDat/pull/173, sorry for that.

I will have another look at this PR now.

static/pdf/comphaus_copresentable.tex
ScriptRaccoon: Found it @dschepler 

```
See <a href="https://ncatlab.org/nlab/show/Sheaves+in+Geometry+and+Logic" target="_blank">Mac Lane & Moerdijk</a>, Thm. IV.4.2.
```

erm, latex code, you know what I mean

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,109 @@
...
+		
+		Since $\prod_{i \in \I} X_i$ is compact, that implies that the intersection of all $E_f$ is non-empty. But that intersection is precisely $\lim_{i \in \I} X_i$.
+	\end{proof}
+\end{lemma}
+
+\begin{lemma}
+	Suppose we have a cofiltered limit $X = \lim_{i\in \I} X_i$ in $\mathbf{Top}$, or in a full subcategory of $\mathbf{Top}$. Then the canonical subbasis of the topology on $X$, which is the weak topology for the projections $p_i : X \to X_i$, is closed under finite intersections. Thus, it agrees with the canonical basis of the topology on $X$.
```
ScriptRaccoon:
```suggestion
	Suppose we have a cofiltered limit $X = \lim_{i\in \I} X_i$ in $\mathbf{Top}$. Then the canonical subbasis of the topology on $X$, which is the weak topology for the projections $p_i : X \to X_i$, is closed under finite intersections. Thus, it agrees with the canonical basis of the topology on $X$.
```

... because limits in a full subcategory don't need to look as expected, and this is not what we want either. We want to analyse only the **Top**-limit here.

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,109 @@
...
+	
+	\begin{proof}
+		Suppose we have a finite collection of subbasic open sets of the form $U_a = p_{i_a}^{-1}(V_a)$, $a \in \{ 1, \ldots, b \}$, where each $V_a$ is an open subset of $X_{i_a}$. Take a cone $(j, f_a : j \to i_a)$ of the objects $i_1, \ldots, i_b$. We then have
+		$$\bigcap_{a=1}^b U_a = p_j^{-1} \left( \bigcap_{a=1}^b f_a^{-1}(V_a) \right),$$
+		where the right hand side is again in the canonical subbasis.
```
ScriptRaccoon: I think it's redundant to introduce indices for the indices. We can just use a finite subset of the indices (similar to my edit of lemma 1). Also, let's be a bit more precise and write $X_f$ and not $f$ when we mean $X_f$.


```suggestion
		Suppose we have a finite collection of subbasic open sets of the form $U_j = p_j^{-1}(V_j)$, $j \in J$, where $J$ is finite and each $V_j$ is an open subset of $X_j$. Take a cone $(i, f_j : i \to j)$ of the objects $j \in J$. We then have
		$$\bigcap_{j \in J} U_j = p_i^{-1} \left( \bigcap_{j \in J} X_{f_j}^{-1}(V_j) \right),$$
		where the right hand side is again in the canonical subbasis.
```

dschepler: I guess technically speaking the original collection of subbasic open sets could have had multiple entries with the same index in $\mathcal{I}$. Then to apply the revised argument you would have to have an extra step of consolidating such sets into their intersection.

Maybe it might look better if I replace $a$ with $m$ and $b$ with $n$?

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,109 @@
...
+		
+		To summarize, we have shown that for each $n \in \INnz$, there exists $i \in \I$ and a finite cover $\{ V_\lambda \mid \lambda \in \Lambda \}$ of $\im(p_i)$ such that whenever $p_i(x), p_i(y) \in V_\lambda$ for some $\lambda$, we have $|\varphi(x) - \varphi(y)| < 1/n$. Now choose such a $i_n$ and associated finite cover of $\im(p_{i_n})$ for each $n$. Then use the fact that $\I$ is $\aleph_1$-cofiltered to take a cone $(j, f_n : j \to i_n)$ of the objects $i_n$. We then see that whenever we have $x, y\in X$ with $p_j(x) = p_j(y)$, then $\varphi(x) = \varphi(y)$. Thus, $\varphi$ induces a well-defined function on the image of $p_j$. This function is continuous, since by construction for any $n\in \INnz$ and $x \in X$, there is a neighborhood $V$ of $p_j(x)$ such that whenever $p_j(y)\in V$ as well, $|\varphi(y) - \varphi(x)| < 1/n$. By the Tietze extension theorem, this induced function can then be extended to a continuous function $\psi : X_j \to [0,1]$. Then $\varphi = \psi \circ p_k$.
```
ScriptRaccoon: Last sentence: `Then $\varphi = \psi \circ p_j$.`

static/pdf/comphaus_copresentable.tex
```diff
@@ -0,0 +1,109 @@
...
+	\end{proof}
+\end{proposition}
+
+This shows that $\CompHaus^{\op}$ is equivalent to the category of algebras over the monad $S \mapsto \Hom_{\CompHaus}([0, 1]^S, [0, 1])$. We may view such algebras as being models of the infinitary algebraic theory of all continuous functions $[0,1]^S \to [0,1]$. In fact, we can show that any such function only depends on countably many coordinates in the domain, so that operations of this theory will be generated by the continuous functions $[0,1]^\omega \to [0,1]$. Indeed, we get the following somewhat stronger result:
+
+\begin{proposition}
+	The object $[0,1]$ of $\CompHaus$ is $\aleph_1$-copresentable. \cite{GU71}
```
ScriptRaccoon:
```suggestion
	The object $[0,1]$ of $\CompHaus$ is $\aleph_1$-copresentable.
```

... because the citation looks like this result is proven there, but I assume it's not

Or, did you want to cite a different article?

static/pdf/comphaus_copresentable.tex
ScriptRaccoon: > I guess technically speaking the original collection of subbasic open sets could have had multiple entries with the same index in $I$

Not really. $p_i^{-1}$ is compatible with intersections and unions.

> Maybe it might look better if I replace a with m and b with n ?

Something like that, yes. I really have a thing against $a \in \\{1,\dotsc,b\\}$ I must say.

static/pdf/comphaus_copresentable.tex
dschepler: The introduction of the paper I got the GU71 citation from said that result was proved there.

dschepler:
Going back and rereading the introduction to one of the cited papers prompted me to refine the last paragraph of the corollary proof a bit:

We thus reproduce a result from \cite{Isb82} which also provides a nice description of a small set of generators of the operations of the $\aleph_0$-ary algebraic theory. A more recent treatment in \cite{Marra_2017} refines this by providing a nice axiomatization of the relations of that theory.

static/pdf/comphaus_copresentable.tex
ScriptRaccoon: Ah, I didn't know that this book goes into topological details. So I checked all results for compact Hausdorff spaces. (In that book, "Hausdorff" is included in "compact".)

- 4.7 proves that $\\{I \times I\\}$ is dense in $CompHaus^{op}$ (which was first shown by Isbell)
- 4.8 proves that $CompHaus$ has no small dense subcategory.
- 6.5 (a) proves that in $CompHaus$ only the empty space is generated (in the sense of $\kappa$-generated for some cardinal $\kappa$).
- 6.5. (b) proves that in $CompHaus^{op}$, a space is finitely generated <=> finitely presented <=> finite.
- 6.5 (c) proves that $I$ is $\aleph_1$-presentable in $CompHaus^{op}$.
- 9.4 (d) proves that in $CompHaus^{op}$, a space is $\aleph_1$-generated <=> $\aleph_1$-presentable <=> metrizable. (The proof reduces to the fact for $I$.)

I can translate the proof of 6.5 (c) and post it here, if you like. It is quite different. In any case, let's add the more precise reference.

dschepler: I guess I ended up reproving 6.5(a) in the proof that CompHaus is not accessible (which is no longer needed now that we know it's locally copresentable).

ScriptRaccoon (approved):

static/pdf/comphaus_copresentable.tex
ScriptRaccoon: I still would prefer writing $X_f$ instead of $f$, though. But I let you decide here.

ScriptRaccoon:
Good work :)

What about the category of locally compact Hausdorff spaces? :D

ScriptRaccoon:
Remark: after merging of #184 the pdf from this PR has now been removed and replaced with the markdown-rendered page https://catdat.app/content/comphaus_copresentable

Issue #91, opened 2026-04-15
Decide accessible etc. for various categories


ScriptRaccoon:
In #59 (resp. #97) several new properties of categories have been added to the database: multi-complete (and the dual), multi-terminal object (and the dual), locally copresentable, accessible (and the dual), ℵ₁-accessible (NOT the dual), finitely accessible (NOT the dual), generalized variety, multi-algebraic, locally multi-presentable, locally finitely multi-presentable, locally poly-presentable. (Also other properties which however are OK.)

The PR did not decide these properties for several categories. But this should be done, in particular for those categories where - before that PR - all properties had been decided. (Let us ignore difficult categories like the category of schemes for now.) 

Namely, for these categories we should decide the mentioned properties. A first PR can handle only the easy cases. A later PR can handle the difficult cases. Or we can have one PR per property. In either case: PRs should be small, something like #59 (resp. #97) should not happen again.

- [Sp](http://localhost:5173/category/Sp)
- [FinGrp](http://localhost:5173/category/FinGrp)
- [FinOrd](http://localhost:5173/category/FinOrd)
- [B](http://localhost:5173/category/B)
- [FI](http://localhost:5173/category/FI)
- [FS](http://localhost:5173/category/FS)
- [Ab_fg](http://localhost:5173/category/Ab_fg)
- [FreeAb](http://localhost:5173/category/FreeAb)
- [Met](http://localhost:5173/category/Met)
- [Setne](http://localhost:5173/category/Setne)
- [Man](http://localhost:5173/category/Man)
- [BG_f](http://localhost:5173/category/BG_f)
- [BG_c](http://localhost:5173/category/BG_c)
- [BN](http://localhost:5173/category/BN)
- [walking_fork](http://localhost:5173/category/walking_fork)
- [walking_pair](http://localhost:5173/category/walking_pair)
- [Delta](http://localhost:5173/category/Delta)

Ignore the properties that are unrelated to the PR (for example if **Met** is regular).

ScriptRaccoon:
@ykawase5048 has mentioned that he can do this. :)

ykawase5048:
It seems that Grp and Set^op should not be included in the list.

ScriptRaccoon:
I meant FinGrp, fixed.
Yes Set^op works now.

ScriptRaccoon:
Remark: I have added a section on small PRs

[...]

PR #179, opened 2026-05-14
Add property: ℵ₁-filtered colimits

ScriptRaccoon:
This PR adds two properties:

- has ℵ₁-filtered colimits
- is ℵ₁-filtered

As for being ℵ₁-filtered, all categories have been decided. For all categories except **Sch**, the existence of ℵ₁-filtered colimits has been decided.

For the categories **Man**, **Grp**<sub>c</sub>, and **Set**<sub>c</sub>, the database already contained proofs that they are not ℵ₁-accessible, but these proofs in fact showed that ℵ₁-filtered colimits do not exist. These were quick wins.

To decide the existence of ℵ₁-filtered colimits for various small categories, and to determine their ℵ₁-accessibility, the following theorem has been added together with its (long) proof:

> **Theorem.** If a category is locally finite, essentially countable, and Cauchy complete, then it is ℵ₁-accessible.

In particular, ℵ₁-filtered colimits exist in this case. The theorem implies ℵ₁-accessibility of **FinOrd**, $\Delta$, **FI**, **FS**, and **B** (which was previously open), as well as **FinSet**, **FinGrp**, and **FinAb** (for which proofs already existed, albeit imprecise ones).

This PR also adds a proof that **Sp** is  ℵ₁-accessible, that **BN** is ℵ₁-accessible and that **BOn** has ℵ₁-filtered colimits (although it is not accessible since it is not well-powered). Questions on mathoverflow have shown that **FreeAb** and **Met**<sub>c</sub> do not have ℵ₁-filtered colimits. In particular, these categories are not ℵ₁-accessible, which has been open before. That being said, this PR already does some of the work for #91.

The only remaining category is **Sch**, the category of schemes. As of writing this, CatDat does not even record whether it has filtered colimits (presumably not, otherwise ind-schemes would not be a thing). Deciding whether it has ℵ₁-filtered colimits has therefore been postponed, as has already been done for several other properties of this category.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -64,6 +64,7 @@
 		"coexponentials",
 		"cofiltered",
 		"cofiltering",
+		"cofinal",
 		"cofinitary",
 		"cofull",
 		"cogenerates",
@@ -125,6 +126,7 @@
 		"extremal",
 		"fieldification",
 		"finitary",
+		"foundedness",
 		"Freyd",
 		"functor",
 		"functorial",
@@ -265,5 +267,8 @@
 		".netlify",
 		"build"
 	],
-	"ignoreRegExpList": ["\\$[^$]*\\$", "\\$\\$[^$]*\\$\\$"]
+	"ignoreRegExpList": [
+		"\\$[^$]*\\$",
+		"\\$\\$[^$]*\\$\\$"
+	]
 }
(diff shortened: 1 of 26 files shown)


ScriptRaccoon:
@ykawase5048  Can you perhaps help here?

Issue #182, opened 2026-05-15
Strengthen additive_effective_congruences_imply_normal


dschepler:
On analysis of one of the explicit assignments in **FreeAb**, we discovered that the implication additive_effective_congruences_imply_normal can be strengthened by removing the assumption of having pullbacks. The revision to be made is:
Given $i : Y \hookrightarrow X$, form the congruence of $\equiv \pmod{i(Y)}$ as $E := X \times Y$, with one map being $(x, y) \mapsto x+i(y)$ and the other map being $(x,y) \mapsto x$.

PR #185, opened 2026-05-15
Strengthen additive_effective_congruences_imply_normal

dschepler:
Addresses #182


ScriptRaccoon:
Nice.

The implication is now used in 6 proofs. Before, that number was 5. In other words, this change had *only* an effect on **FreeAb**.

```sql
select count(*) from category_property_assignments where is_deduced = TRUE and reason like '%category-implicatione_effective_congruences_imply_normal%';
```

ScriptRaccoon (approved):

PR #186, opened 2026-05-16
Add Barr-(co)exact properties

dschepler:
Adds Barr-exact and Barr-coexact properties, and makes a few adjustments to existing category property assignments accordingly. This should also make it easy to add a "pretopos" property.

### DIFF
--- a/databases/catdat/data/categories/Set_pointed.yaml
+++ b/databases/catdat/data/categories/Set_pointed.yaml
@@ -35,6 +35,7 @@ satisfied_properties:
 
   - property: coregular
     reason: From the other properties we know that (co-)limits exist and that monomorphisms coincide with injective pointed maps. So it suffices to prove that these maps are stable under pushouts. This follows from the corresponding fact for <a href="/category/Set">$\Set$</a> and the observation that the forgetful functor $\Set_* \to \Set$ preserves pushouts.
+    check_redundancy: false
 
   - property: co-Malcev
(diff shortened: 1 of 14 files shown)


databases/catdat/data/categories/Set_pointed.yaml
```diff
@@ -35,6 +35,7 @@ satisfied_properties:
 
   - property: coregular
     reason: From the other properties we know that (co-)limits exist and that monomorphisms coincide with injective pointed maps. So it suffices to prove that these maps are stable under pushouts. This follows from the corresponding fact for <a href="/category/Set">$\Set$</a> and the observation that the forgetful functor $\Set_* \to \Set$ preserves pushouts.
+    check_redundancy: false
```
ScriptRaccoon, line 38: Why do we keep this assignment?

ScriptRaccoon:
Thank you! I will have a look at it later. For now I have just skimmed through it and it looks fine. I only added a minor comment.

ScriptRaccoon (approved):

databases/catdat/data/categories/Set_pointed.yaml
dschepler: I was actually debating whether to keep it or not, and decided we might want to keep it since it's a simpler proof than the one using the heavy machinery of monadic functors and slices.

That said, I could certainly agree if you want to remove this and rely on the Barr-coexact assignment.

PR #210, opened 2026-05-27
Sh(X) never has a generator (unless X has the indiscrete topology)

dschepler:
It turns out this only resolves the direct assignment plus not being finitary algebraic.

---

- unknown (category, property)-pairs before this PR: 116
- unknown (category, property)-pairs after this PR: 114

### DIFF
--- a/databases/catdat/data/categories/Sh(X).yaml
+++ b/databases/catdat/data/categories/Sh(X).yaml
@@ -16,7 +16,7 @@ related_categories:
   - Sh(X,Ab)
 
 comments:
-  - It is likely that neither of the currently remaining unknown properties (finitary algebraic, locally ℵ₁-presentable, exact filtered colimits, etc.) are satisfied for a <i>generic</i> space $X$, but we need to make this precise by adding additional requirements to $X$. Maybe we need to create separate entries for specific spaces $X$.
+  - It is likely that none of the currently remaining unknown properties (locally ℵ₁-presentable, exact filtered colimits, etc.) are satisfied for a <i>generic</i> space $X$, but we need to make this precise by adding additional requirements to $X$. Maybe we need to create separate entries for specific spaces $X$.
 
 satisfied_properties:
   - property: locally small
@@ -29,8 +29,11 @@ unsatisfied_properties:
   - property: skeletal
     reason: Consider constant sheaves for isomorphic but non-equal sets.
 
-  - property: trivial
-    reason: This is because $X$ is assumed to be non-empty.
+  - property: generator
+    reason: >-
+      Suppose $\Sh(X)$ had a generator. Then by <a href="/content/topos-with-generator">this result</a>, every subterminal object would be either initial or terminal. On the other hand, the subterminal objects of $\Sh(X)$ are of the form $y_U \coloneqq \Hom({-}, U)$ for $U$ open. (Alternative descriptions of $y_U$ include:
+      $$y_U(V) = \begin{cases} 1, & V \subseteq U; \\ \varnothing & \mathrm{otherwise} \end{cases}$$
+      with the unique restriction maps; and $y_U$ represents the functor $\Gamma(U, {-})$ of sections over $U$.) Therefore, the lemma implies that every open set of $X$ is either $\varnothing$ or $X$, contradicting our assumption that $X$ does not have the indiscrete topology.
 
 special_objects:
   initial object:
--- a/.cspell.json
+++ b/.cspell.json
@@ -252,6 +252,7 @@
 		"subscheme",
 		"subsemigroup",
 		"subsheaf",
+		"subterminal",
 		"summands",
 		"suprema",
 		"supremum",
--- a/content/topos-with-generator.md
+++ b/content/topos-with-generator.md
@@ -1,0 +1,21 @@
+---
+title: Topos with a Generator
+description: An elementary topos with a generator has at most two subterminal objects
+author: Daniel Schepler
+---
+
+# Topos with a Generator
+
+::: Lemma
+Suppose a category is coregular, and it has disjoint finite coproducts, a terminal object, and a generator. Then every regular subterminal object (i.e. an object $X$ such that the unique morphism $X \to 1$ is a regular monomorphism) is either initial or terminal.
+:::
+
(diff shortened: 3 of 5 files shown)


ScriptRaccoon:
Maybe we should even assume that $X$ is sober (a space has the same category of sheaves as its soberification, right?).

If you agree, you can add this right away to the file. Then this weird "not indiscrete" assumption becomes redundant, I assume.

databases/catdat/data/categories/Sh(X).yaml
```diff
@@ -32,6 +32,9 @@ unsatisfied_properties:
   - property: trivial
     reason: This is because $X$ is assumed to be non-empty.
 
+  - property: generator
+    reason: Since we are assuming $X$ is not indiscrete, let $U$ be an open set which is neither empty nor all of $X$. Then any generator $G$ must have a morphism to $y_U$ in order to distinguish the two inclusion morphisms $y_U \to y_U + y_U$. (Here $y_U := \Hom({-}, U)$ is the sheaf where $y_U(V) = 1$ if $V\subseteq U$, or $y_U(V) = \varnothing$ otherwise, which represents the functor $\Gamma(U, {-})$ of sections over $U$.) That implies that $G(V)$ can be non-empty only for $V \subseteq U$. However, that contradicts the fact that $G$ must also distinguish the two inclusion maps $1 \to 1 +_{y_U} 1$.
```
ScriptRaccoon:
```suggestion
    reason: Since we are assuming $X$ is not indiscrete, let $U$ be an open set which is neither empty nor all of $X$. Then any generator $G$ must have a morphism to $y_U$ in order to distinguish the two inclusion morphisms $y_U \rightrightarrows y_U + y_U$. (Here $y_U \coloneqq \Hom({-}, U)$ is the sheaf where $y_U(V) = 1$ if $V\subseteq U$, or $y_U(V) = \varnothing$ otherwise, which represents the functor $\Gamma(U, {-})$ of sections over $U$.) That implies that $G(V)$ can be non-empty only for $V \subseteq U$. However, that contradicts the fact that $G$ must also distinguish the two inclusion maps $1 \rightrightarrows 1 +_{y_U} 1$.
```

ScriptRaccoon: Can you add a proof that

1. the two $y_U  \rightrightarrows y_U + y_U$ are in fact different*,
2. the two $1 \rightrightarrows 1 +_{y_U} 1$ are in fact different*,
3. how the contradiction is happening in the end exactly

*not entirely trivial since the sheaf colimit is the associated sheaf to the presheaf colimit, or am I missing something?

Also, can we turn this proof into an implication ("elementary topos + generator => ...)? I think I have recently read that it must be 2-valued. Hmm, except we don't have this property right now.

dschepler: For the first two, I could either make use of the fact that stalks commute with coproducts/pushouts, or make use of the fact that the presheaf coproducts/pushouts are separated, and the canonical map from a separated presheaf to its sheafification is a monomorphism. For the third, the contradiction comes from the fact that the pushout has singleton sections over any $V \subseteq U$. I'll think a bit on how best to word that.

dschepler: Hmm, I guess generalizing the argument to use any subterminal object of an elementary topos in place of $y_U$ could very well show that the only subterminal objects are 0 and 1.  If the two maps $S \rightrightarrows S + S$ are equal, then disjointness of coproducts implies that $S \simeq 0$. And if the two maps $1 \rightrightarrows 1 +_S 1$ are equal, then a simple internal language argument, using the construction of the pushout as the quotient of $1 + 1$ by the diagonal unioned with $x \sim x'$ and $x' \sim x$ for $x : S$, should show that $S \simeq 1$.  And then, if both pairs are unequal, you get a contradiction from the fact that the unique morphism $G \to 1$ must factor through $S$ from the first, but then there's only one morphism $G \to 1 +_S 1$.

If you like, I could try to write that up as a new lemma, which could then be transformed into an implication if you do decide to add a property 2-valued, assuming that means having only 0 and 1 as subterminal objects.

ScriptRaccoon: Sounds good!

I think we can also say that the equalizer of $S \rightrightarrows S + S$ is $0$ (as you say: disjoint coproducts) and that the equalizer of $1 \rightrightarrows 1+_S 1$ is $S$ (every monomorphism is effective; it would be good to have a reference for this). I think this makes the proof more clear.

dschepler: Actually, I guess the equalizer of $X \rightrightarrows X +_Y X$ for $Y \hookrightarrow X$ monic being $Y$ would be a consequence of the elementary topos being coregular and mono-regular.

dschepler: I suppose that would also give a proof of: coregular + mono-regular + strict initial object -> disjoint binary copowers - except we don't have such a property at the moment.

(I've also been thinking of adding an implication: regular + balanced -> epi-regular. Proof: in the regular epi - mono factorization of an epimorphism, the mono part must also be epi, so it's an isomorphism. The dual of that would give an automatic deduction of coregular + balanced + strict initial object -> disjoint binary copowers.)

dschepler:
The new lemma could potentially be used to rewrite the assignments for SetxSet (where $(1,0)$ and $(0,1)$ are each a subterminal object which is not initial or terminal) and for Sp (where I guess for any $n$ the functor assigning 1 for sets of cardinality $n$ and $\varnothing$ otherwise is an intermediate subterminal object). Not sure whether it's worth it for SetxSet (in fact, the proof is easier in general in any situation where the intersection of all intermediate subterminal objects is initial) but in the existing argument for Sp I do see echoes (or foreshadowing) of the general lemma's proof.

databases/catdat/data/categories/Sh(X).yaml
ScriptRaccoon: Thanks for the revision!
I think these implication(s) can be added in a separate PR.

ScriptRaccoon (approved):
Looks very good! Can you please squash your commits into one? Then I will merge.

ScriptRaccoon:
> The new lemma could potentially be used to rewrite the assignments for SetxSet [...] and for Sp

Yes. But let's only *add* a sentence like "Alternatively, this follows from ..."

dschepler:
> Looks very good! Can you please squash your commits into one? Then I will merge.

Done.

PR #212, opened 2026-05-27
Add the Crude Monadicity Theorem

ScriptRaccoon:
- remove incorrect implication `full + monadic => equivalence` and decide fullness manually (this result has even caused wrong assignments, e.g. that the discrete topology functor is not full)
- add new properties of functors: preserves **reflexive coequalizers** / preserves **coreflexive equalizers**
- decide these properties for all functors in the database
- add the [**crude monadicty theorem**](https://ncatlab.org/nlab/show/monadicity+theorem#CrudeMonadicityTheorem)
- refactor existing assignments of functors: instead of citing a result that shows monadicity, show that the assumptions of crude monadicity are fulfilled (only one exception: $X \mapsto X^\omega$ on $Set$)
- add the little result that on a mono-regular category, a functor preserving equalizers also preserves monomorphisms; remove assignments based on this
- other minor improvements


databases/catdat/data/functor-implications/monadic.yaml
ScriptRaccoon, line 1: absolute useless diff here :D

PR #217, opened 2026-05-28
Add missing proofs for FinGrp, FS, Met

ScriptRaccoon:
This PR 

- adds a proof that **FinGrp** has effective cocongruences (proven in https://mathoverflow.net/questions/511516)
- adds a proof that **FS** is a generalized variety
- greatly simplifies the proof that **FS** is finitely accessible
- adds a proof that **PMet**, **Met** and **Met**<sub>&infin;</sub> are not regular

---

- unknown (category, property)-pairs before this PR: 114
- unknown (category, property)-pairs after this PR: 108


content/sifted-colimits-in-groupoids.md
```diff
@@ -6,14 +6,18 @@ author: Martin Brandenburg
 
 ## Sifted colimits in groupoids
 
-While the combination of [this result](http://localhost:5173/category-implication/groupoid_consequence) and [this result](http://localhost:5173/category-implication/sifted_colimits_criterion) already implies that groupoids have sifted colimits, we can make these colimits more explicit and also prove their existence without using any non-trivial theorem. We also generalize the situation as follows.
+While the combination of [this result](http://localhost:5173/category-implication/groupoid_consequence) and [this result](http://localhost:5173/category-implication/sifted_colimits_criterion) already implies that groupoids have sifted colimits, we can make these colimits more explicit and also prove their existence without using any non-trivial theorem. We also do this in a more general setting.
+
+Let $D : \I \to \C$ be a diagram in a category and $i_0 \in \I$. We call $D$ _constant after_ $i_0$ when for all morphisms $i_0 \to i$ the morphism $D(i_0) \to D(i)$ is an isomorphism. Of course, in a groupoid, this is satisfied for every $i_0 \in \I$. If such an object $i_0$ exists, we call $D$ _eventually constant_.
```
dschepler: I'm thinking that leaving open the possibility to call a non-sifted diagram satisfying this criterion "eventually constant" could be highly misleading. For example, in the mere "diagram with cospans" of a parallel pair $\{ 1, 2 \} \rightrightarrows \{ 1, 2 \}$ where one map is the identity and the other is the transposition, every morphism is an isomorphism but the coequalizer is a singleton.

I was thinking of the criterion as being "the diagram eventually consists of isomorphisms" (or less formally "the diagram eventually settles down to isomorphisms"), and it's the sifted criterion that forces those eventual isomorphisms to be consistent with each other and with the other morphisms.

Of course, this could be considered a minor quibble and it's up to you whether to change the terminology.

dschepler:
Other than the one minor reservation I have about the terminology "eventually constant", this looks good to me.

content/sifted-colimits-in-groupoids.md
ScriptRaccoon: Thanks. Which formulation of that section do you suggest, exactly? (I am unsure if in the 2nd paragraph of your comment the restriction to sifted diagram still stands.)

dschepler: I really don't know what exactly I would suggest, unfortunately.

ScriptRaccoon: I just added the restriction that D is sifted. I hope this is OK.

PR #225, opened 2026-06-02
LRS does not have a small generating set

dschepler:
This took the number of unknown properties of LRS from 27 to 16.

---

- unknown (category, property)-pairs before this PR: 108
- unknown (category, property)-pairs after this PR: 97

### DIFF
--- a/databases/catdat/data/categories/LRS.yaml
+++ b/databases/catdat/data/categories/LRS.yaml
@@ -48,7 +48,10 @@ unsatisfied_properties:
     proof: 'We can adjust the proof for <a href="/category/Top">$\Top$</a> (see <a href="https://mathoverflow.net/questions/509548" target="_blank">MO/509548</a>) as follows: Let $k$ be a field, $X$ be a singleton and $Y = \{u,v\}$ be the Sierpinski space where $\{u\}$ is open, but $\{v\}$ is not. Endow both with the sheaf of locally constant functions to $k$. Thus, $\O_X(X) = k$, $\O_Y(Y) = \O_Y(\{u\}) = k$. There is a canonical morphism $p : X + X \to Y$. It is a coreflexive corelation that is not cosymmetric.'
 
   - property: generating set
-    proof: Out of any small set $S$ of locally ringed spaces, there is only a small set of residue fields. Therefore, if $k$ is a field of characteristic $\ne 2$ with a strictly larger cardinality than any of these residue fields, then the only possible morphism from an element of $S$ to $\Spec k(t)$ is one with an empty domain. However, that makes it impossible for $S$ to distinguish the two morphisms $\Spec k(t) \to \Spec k(t)$ corresponding to the two automorphisms in $\Gal(k(t) / k(t^2))$ (i.e. the identity automorphism and the automorphism which sends a rational function $f(t) \mapsto f(-t)$).
+    proof: >-
+      Out of any small set $S$ of locally ringed spaces, there is only a small set of residue fields at their points. Therefore, if $k$ is a field of characteristic $\ne 2$ with a strictly larger cardinality than any of these residue fields, then the only possible morphism from an element of $S$ to $\Spec k(t)$ is one with an empty domain. However, that makes it impossible for $S$ to distinguish the two morphisms $\Spec k(t) \rightrightarrows \Spec k(t)$ induced by $t \mapsto \pm t$.
+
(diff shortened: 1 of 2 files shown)


databases/catdat/data/categories/LRS.yaml
```diff
@@ -47,6 +47,9 @@ unsatisfied_properties:
   - property: co-Malcev
     proof: 'We can adjust the proof for <a href="/category/Top">$\Top$</a> (see <a href="https://mathoverflow.net/questions/509548" target="_blank">MO/509548</a>) as follows: Let $k$ be a field, $X$ be a singleton and $Y = \{u,v\}$ be the Sierpinski space where $\{u\}$ is open, but $\{v\}$ is not. Endow both with the sheaf of locally constant functions to $k$. Thus, $\O_X(X) = k$, $\O_Y(Y) = \O_Y(\{u\}) = k$. There is a canonical morphism $p : X + X \to Y$. It is a coreflexive corelation that is not cosymmetric.'
 
+  - property: generating set
+    proof: Out of any small set $S$ of locally ringed spaces, there is only a small set of residue fields. Therefore, if $k$ is a field of characteristic $\ne 2$ with a strictly larger cardinality than any of these residue fields, then the only possible morphism from an element of $S$ to $\Spec k(t)$ is one with an empty domain. However, that makes it impossible for $S$ to distinguish the two morphisms $\Spec k(t) \to \Spec k(t)$ corresponding to the two automorphisms in $\Gal(k(t) / k(t^2))$ (i.e. the identity automorphism and the automorphism which sends a rational function $f(t) \mapsto f(-t)$).
```
ScriptRaccoon: You mean the set of all residue fields of all spaces in S I assume?

Ar first I thought you mean the spaces in S that have the form spec(k).

ScriptRaccoon:
> This took the number of unknown properties of LRS from 27 to 16.

Great progress! 🙏🏻

ScriptRaccoon (approved):

databases/catdat/data/categories/LRS.yaml
dschepler: Yes, I mean the set of residue fields of any space in S at any point.

ScriptRaccoon:
Here is an alternative proof.

Lemma: Let C be a full reflective subcategory of D. Assume that D has a generating set. Then C has a generating set. (Namely, the reflector applied to a generating set.)

Apply this to the category of affine schemes which is a reflective category of the category of locally ringed spaces. The reflector maps X to the spectrum of the ring of global sections of X. But the category of affine schemes has no generating set since CRing has no cogenerating set.

This also provides an alternative proof for the category of schemes.

(I hope this is correct. I am not fully concentrating right now and will check the details later.)

ScriptRaccoon:
Actually, your proof heavily reminds me of the lemma on Missing cogenerating sets which we already use for categories like CRing. Maybe it applies (dualized) by taking F to be the collection of spectra of fields. We only need to define U.

dschepler:
I guess the functor of taking a space to its collection of residue fields could be viewed as a functor $\mathbf{LRS} \to \mathbf{Set}^\rightarrow$ where $X$ maps to the morphism with codomain given by the underlying set of $X$ and domain given by the disjoint union of (the underlying sets of) the residue fields. And for morphisms $f : X \to Y$ we get an injective map from the pullback of $U(Y)$ and the underlying set map of $f$ to the domain of $U(X)$ -- so I guess the target category is not quite $\mathbf{Set}^\rightarrow$ but something with modified morphisms. And even though $U$ might be faithful in some sense, we don't actually need the full power of faithfulness of $U$, just that it acts faithfully on endomorphisms of the objects in $\mathbf{F}$ -- for which in this case, $U$ on any object in $\mathbf{F}$ has a singleton codomain.

But anyway, all that setup seems like a lot of unnecessary complication to generalize the lemma just for this one case.

Edit: A bit of Google searching gave a name of the target category, such that the functor described is a contravariant functor from $\mathbf{LRS}$ to "the total category of the codomain fibration over $\mathbf{Set}$".  And I was reminded this is not a faithful functor in general (for example looking at the endomorphisms of $\mathbb{C}[\varepsilon] / \langle \varepsilon^2 \rangle$), so the refinement to use faithfulness on endomorphisms of objects in $\mathbf{F}$ is important in this case.

ScriptRaccoon:
Alright. I have made a commit where I have adjusted your proof slightly, and where I also added the alternative proof using the adjunction. Can you please check if it is OK for you?

dschepler:
Yes, the edits you made look good to me.

ScriptRaccoon:
<s>I just noticed that it's even easier to prove that LRS and Sch don't have a cogenerating set. PR will follow soon.</s>

I made a mistake.

PR #231, opened 2026-06-06
Add several negative properties of LRS_R deduced from those of Top

### LINKED ISSUE #227: What to do with my proof that LRS is not cartesian closed?
Hmm...  I was planning to submit a proof I've found that LRS is not cartesian closed.  The basic idea was to prove that if k is a finite field, then LRS_k is not cartesian closed.  The idea of the proof is to examine the functor $\mathbf{Top} \to \mathbf{LRS}_k$ of equipping a topological space with the constant sheaf of $k$.  That has a left adjoint given by the forgetful functor, and it has a right adjoint taking a locally ringed space over k to the subspace of points where the canonical morphism from k to the residue field is an isomorphism.  The proof of the right adjoint part, in particular, requires the finiteness assumption on k in order to construct locally constant functions.  Then, I would use this general lemma to get a contradiction: if D is a coreflective subcategory of C which is essentially closed under binary products (or equivalently D has binary products and the inclusion functor preserves them), and C is cartesian closed, then D is cartesian closed.

Then, to pass to LRS not being cartesian closed, I would use a lemma: if C is a cartesian closed category, and P is a subterminal object, then the slice C/P is also cartesian closed.  So, if LRS were cartesian closed, then LRS/F_p would be cartesian closed also for each prime p, giving a contradiction.

But then, I noticed you just pushed changes replacing LRS with LRS_R and Sch with Sch_R.  So now, I'm not sure how to proceed since my proof (at least at the moment) is limited to rings R which have a finite field as a quotient.

dschepler:
Addresses: #227

### DIFF
--- a/content/Top-embeds-in-LRS.md
+++ b/content/Top-embeds-in-LRS.md
@@ -1,0 +1,52 @@
+---
+title: An embedding of the category of topological spaces in the category of locally ringed spaces
+description: Describes a functor which makes the category of topological spaces a coreflective and "almost reflective" subcategory of the category of locally ringed spaces. From the properties of this embedding, we can rule out several properties for the category of locally ringed spaces, using the corresponding failures of these properties for the category of topological spaces.
+author: Daniel Schepler
+---
+
+## An embedding of the category of topological spaces in the category of locally ringed spaces
+
+For much of this development, we will be dealing with the case of $\LRS_k$ where $k$ is a field. We begin by describing $\Top$ as a reflective subcategory of $\LRS_k$.
+
+::: Lemma 1
+The forgetful functor $U : \LRS_k \to \Top$ has a right adjoint $K : \Top \to \LRS_k$ of equipping a topological space $X$ with the constant sheaf $\underline{k}$. Furthermore, the functor $K$ is fully faithful, thus making $\Top$ into a reflective subcategory of $\LRS_k$.
+:::
+
+_Proof._ In this adjunction, the counit $UK \to \id$ is just the identity. To describe the unit $\id \to KU$, we need to define a morphism $(X, \O_X) \to (X, \underline{k})$ for any locally ringed space $(X, \O_X)$ over $k$. This morphism will be the identity on topological spaces, and the pullback operation $\underline{k} \to \O_X$ will be the unique morphism of sheaves induced by the given structure of $\O_X$ as a sheaf of $k$-algebras. It is now straightforward to check this indeed defines an adjunction; and since the counit is an isomorphism, that implies that $K$ is fully faithful. <span class="qed">$\square$</span>
+
+We now show that this reflective subcategory is in fact also a coreflective subcategory. Recall that for $f \in \O_X(U)$ we have its vanishing set $V(f) \coloneqq \{x\in U : f \in \m_{X,x}\} = \{x \in U : f(x) = 0\}$, where $f(x) \in \kappa(x)$ is the image of $f_x \in \O_{X,x}$ in the residue field.
+
+::: Lemma 2
+For each object $X$ of $\LRS_k$, let $X_0$ be the set of points $x \in X$ such that the induced morphism from $k$ to the residue field $\kappa(x)$ is an isomorphism. We give $X_0$ the following strengthening of the subspace topology: it will be the topology where a neighborhood subbasis at $x \in X_0$ is the collection of sets of the form $X_0 \cap V(f)$ where $f \in \O_X(U)$ for some neighborhood $U$ of $x$ in $X$, and $x \in V(f)$. Then $X \mapsto X_0$ defines a functor $S : \LRS_k \to \Top$ that is right adjoint to $K$.
+:::
+
+_Proof._ First, to see that $S$ is a functor, suppose we have a morphism $f : (X, \O_X) \to (Y, \O_Y)$. Then for $x \in X_0$, we have a sequence $k \to \kappa(f(x)) \to \kappa(x)$ where the composition is an isomorphism. Thus, $\kappa(f(x)) \to \kappa(x)$ is a surjective morphism of fields, and therefore an isomorphism. It follows that $k \to \kappa(f(x))$ is also an isomorphism of fields, so $f(x) \in Y_0$. To see that the restriction map $X_0 \to Y_0$ is continuous, suppose $g \in \O_Y(V)$ is such that $f(x) \in V(g)$. Then $x \in V(f^\sharp g)$, and
+$$X_0 \cap f^{-1}(Y_0 \cap V(g)) = X_0 \cap V(f^\sharp g)$$
+where $f^\sharp g \in \O_X(f^{-1}(V))$. In other words, we have shown that the inverse image in $X_0$ of any subbasic neighborhood of $f(x)$ is a neighborhood of $x$.
+
+Now if we apply the functor $S$ to a space of the form $(X, \underline{k})$, then since by definition any section of the constant sheaf $\underline{k}$ is locally constant, we see that we recover exactly $X$ with its original topology. Thus, we can define the unit $\id \to SK$ of the adjunction to be the identity.
+
+As for the counit $KS \to \id$, for any locally ringed space $(X, \O_X)$ over $k$ we need to define a morphism $(X_0, \underline{k}) \to (X, \O_X)$. The map of topological spaces will be the inclusion map $i : X_0 \hookrightarrow X$, which is continuous since in particular for $U$ an open neighborhood of $x \in X_0$ we have $X_0 \cap U = X_0 \cap V(0_U)$, where $0_U \in \O_X(U)$ is the zero element. The pullback map $\O_X \to i_* \underline{k}$ takes $f \in \O_X(U)$ to the function $X_0 \cap U \to k$ where $x \in X_0 \cap U$ maps to the inverse image of $f(x) \in \kappa(x)$ under the isomorphism $k \to \kappa(x)$. An alternative description of this pullback is that $x \in X_0 \cap U$ maps to the unique $a\in k$ such that $x \in V(f-a)$. Since $X_0 \cap V(f-a)$ is a neighborhood of $x$ in $X_0$ by definition, this shows that we get a locally constant function to $k$ as required.
+
+From here, it is straightforward to show that this does in fact define an adjunction. <span class="qed">$\square$</span>
+
+_Remark._ In the special case where $k$ is a finite field, we have
+$$\textstyle X_0 \cap V(f) = \bigcap_{a \in k^\times} (X_0 \cap D(f-a)),$$
+which is already open in the subspace topology. Therefore, in this case, $X_0$ is given exactly the subspace topology.
+
+::: Corollary 3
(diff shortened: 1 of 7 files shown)


ScriptRaccoon:
Thanks a lot! I am glad that we can finally decide some of the properties of this category. It's also good to see that you could generalize your proof that previously only worked for special ground rings.

I have revised the content page, please check the three 3 commits that I added.

I still need to check the proofs in `LRS_R.yaml`. I will do this in the evening today.

databases/catdat/data/categories/LRS_R.yaml
```diff
@@ -53,6 +53,22 @@ unsatisfied_properties:
 
       Alternatively, using the usual adjunction between affine schemes and locally ringed spaces (<a href="https://en.wikipedia.org/wiki/%C3%89l%C3%A9ments_de_g%C3%A9om%C3%A9trie_alg%C3%A9brique" target="_blank">EGA I</a> (1971), Ch. 1, Prop. 1.6.3), a generating set in $\LRS_R$ would induce a generating set in the category of affine $R$-schemes, which contradicts the fact that <a href="/category/CAlg(R)">$\CAlg(R)$</a> does not have a cogenerating set.
 
+  - property: cartesian closed
+    proof: This is proved <a href="/content/LRS-not-cartesian-closed">here</a>.
+    check_redundancy: false
+
+  - property: cartesian filtered colimits
+    proof: As a corollary of the results <a href="/content/LRS-not-cartesian-closed">here</a>, if we choose a quotient field $k$ of $R$, then the functor $\Top \to \LRS_R$ of equipping a topological space with the constant sheaf of $k$ is fully faithful, and preserves all colimits and all inhabited limits. Therefore, if $\LRS_R$ had cartesian filtered colimits, then $\Top$ would also, giving a contradiction.
```
ScriptRaccoon: I have some issues with this proof and the other ones (but no doubt that they are correct!). It is just said that "of the results here", but not making explicit which results are used for which claim. Also, I think that some of the **proofs** are even used. This suggests refactoring them.

Notice that for example the content page never mentions any functor from Top to LRS<sub>R</sub> where R is a commutative ring which is not a field.

Maybe it is even better to add all these proofs to the content page itself. It can be a huge corollary like "The category LRS_R is (a) not X, (b) not Y, ... ".

Notice that this "outsourcing" has already been done for the cartesian closed property. The current approach is inconsistent.

Also consider adding stuff to the page `cartesian-closed-results.md` that I created for the more general results. Probably it then should also be renamed, since it is not about ccc anymore, but rather about coreflections. Then it can also be merged with `subobject_classifiers_coreflection.md`, and also with `exact_filtered_colimits_descend.md`.

In a later stage of this project, all of this will part of an extended deduction system where properties of categories can be deduced *automatically* from properties and results of (adjoint) functors. Namely, functor implications will then also have fields like `target_conclusions` and `left_adjoint_assumptions`.

ScriptRaccoon (changes requested):

databases/catdat/data/categories/LRS_R.yaml
dschepler: OK, I was thinking of the composition of the functor K : Top -> LRS_k and the forgetful functor LRS_k -> LRS_R. I suppose it would make sense to make that explicit.

I'm not clear on the exact reorganization that you're suggesting. Maybe something like subcategory_inheritance.md which is a catch-all for formal results like "if a subcategory of a regular category is finitely complete and has coequalizers, and the inclusion functor preserves finite limits and coequalizers and is conservative, then the subcategory is regular"? (Though for that particular one, I guess the result wouldn't necessarily need it to be a subcategory inclusion, or even faithful.)

If so, then it might make sense to make that reorganization into a separate PR. Though I guess I could certainly get a start with the general results needed for LRS_R.

Regarding the later stage: I suppose it would also need dependency tracking of the proofs to ensure we don't accidentally end up with circular arguments, right?

ScriptRaccoon: I was only referring to results about (co)reflective subcategories (not general subcategories) and which properties are inherited*. It doesn't need to be complete in any way, just gather the results we are currently using.

Separate PR: I can do this now where I combine `subobject_classifiers_coreflection.md` and `exact_filtered_colimits_descend.md` into, say, `coreflective_subcategories.md`. After merging this, you can rebase and add the results you need there. 

Circular arguments will be prevented, yes, but this is simply done via primary keys in SQL (I think, I will check this more carefully of course). Unfortunately, circular arguments are potentially present already now because of all the "soft references" to other proofs. (Not only in CatDat, but in math as a whole.)

*EDIT: ok actually there is a result which is a bit more general, so yeah, it will be about subcategories in general

ScriptRaccoon: This what I meant: https://github.com/ScriptRaccoon/CatDat/pull/237

dschepler: OK, that looks good.

Regarding the circular arguments: it does make sense that the automated deduction wouldn't create those.  I guess what I was thinking there was along the lines of adding annotations to the "soft references" to make them into checkable references.  (And the redundancy checker could then also take those into account.)  So, something roughly along the lines of:
```
depends:
  - category: Top
    property: cartesian filtered colimits
```

ScriptRaccoon: #237 is merged now.

dschepler:
I have a question about the proof that Top does not have cartesian filtered colimits: the link given is proving that $\mathbb{Q} \times {-}$ does not preserve coequalizers, but coequalizers aren't filtered colimits. Is there some implication involved, not mentioned in the proof, that relates cartesian filtered colimits to "cartesian coequalizers"?

ScriptRaccoon:
> I have a question about the proof that Top does not have cartesian filtered colimits: the link given is proving that Q × − does not preserve coequalizers, but coequalizers aren't filtered colimits. Is there some implication involved, not mentioned in the proof, that relates cartesian filtered colimits to "cartesian coequalizers"?

The link is wrong, it should be https://math.stackexchange.com/questions/1255678 (it was correct once, at some point I made the wrong replacement).

The same link is used for **Haus**.

databases/catdat/data/categories/LRS_R.yaml
```diff
@@ -53,6 +53,22 @@ unsatisfied_properties:
...
+
+  - property: regular
+    proof: This is Corollary 4(c) <a href="/content/Top-embeds-in-LRS">here</a>.
+
+  - property: cofiltered-limit-stable epimorphisms
+    proof: This is Corollary 4(d) <a href="/content/Top-embeds-in-LRS">here</a>.
+
+  - property: effective cocongruences
+    proof: This is Corollary 4(e) <a href="/content/Top-embeds-in-LRS">here</a>.
+
```
ScriptRaccoon, line 71: love it!

content/Top-embeds-in-LRS.md
```diff
@@ -0,0 +1,52 @@
...
+::: Corollary 4
+Let $R$ be any non-trivial commutative ring. Then:<br>
+(a) $\LRS_R$ is not cartesian closed.<br>
+(b) $\LRS_R$ does not have cartesian filtered colimits.<br>
+(c) $\LRS_R$ is not regular.<br>
+(d) $\LRS_R$ does not have filtered-colimit-stable epimorphisms.<br>
+(e) $\LRS_R$ does not have effective cocongruences.
+:::
+
+_Proof._ As before, we fix a quotient field of $R$. In each case, this is an easy application of a contrapositive of a result from [here](/content/subcategories) to the functor $K_R$. Namely, (a) follows from Lemma 5; (b) from the dual of Lemma 4; (c) from Lemma 7; (d) from the dual of Lemma 2 with the observation that $K_R$ preserves epimorphisms since it has a right adjoint; and (e) from the dual of Lemma 8. <span class="qed">$\square$</span>
```
ScriptRaccoon: Please add a sentence like `We know that $\Top$ does not have any of these properties` somewhere.

ScriptRaccoon: I don't understand the proof of (b). Why does the dual of Lemma 4 apply? Wouldn't this be about cofiltered limits? Also, don't we need to check a property of the adjoint?

ScriptRaccoon:
> > I have a question about the proof that Top does not have cartesian filtered colimits: the link given is proving that Q × − does not preserve coequalizers, but coequalizers aren't filtered colimits. Is there some implication involved, not mentioned in the proof, that relates cartesian filtered colimits to "cartesian coequalizers"?
> 
> The link is wrong, it should be https://math.stackexchange.com/questions/1255678 (it was correct once, at some point I made the wrong replacement).

Fixed in https://github.com/ScriptRaccoon/CatDat/commit/275de99faba4693a2b0fe547f15defd1dff15d86

content/Top-embeds-in-LRS.md
dschepler: Whoops, you're right - and it's definitely not the case that the left adjoint U of K preserves binary products. I'll have to think about that some more, and I might end up needing to withdraw the assignment for that property.

content/subcategories.md
```diff
@@ -93,7 +93,7 @@ $$UX \times_{UY} UX \rightrightarrows UX \to \im(Uf) \xrightarrow{Ui} UY$$
 where $UX \times_{UY} UX$ is the kernel pair of $Uf$, and $\im(Uf)$ is the coequalizer. Since $Uf$ is a regular epimorphism, we must have $Ui$ is an isomorphism. Since $U$ is fully faithful and therefore conservative, we get $i$ is an isomorphism as well, so $f$ is a regular epimorphism. <span class="qed">$\square$</span>
 
 ::: Lemma 8
-Let $U : \C \to \D$ be a fully faithful functor. Assume that $\C$ has finite limits and coequalizers, and that $U$ preserves inhabited finite limits and coequalizers. If $\D$ has effective congruences, then so does $\C$.
+Let $U : \C \to \D$ be a fully faithful functor. Assume that $\C$ has finite limits and coequalizers, and that $U$ preserves monomorphisms, inhabited finite limits, and coequalizers. If $\D$ has effective congruences, then so does $\C$.
```
ScriptRaccoon: Any functor preserving pullbacks also preserves monomorphisms. So I suggest to remove this assumption.

ScriptRaccoon (approved):

ScriptRaccoon:
FYI #241

ScriptRaccoon:
The rebase was a bit aggressive 😄 since you now also squashed my commits that were independent from LRS (`src/lib/server/markdown.ts` and `src/routes/content/[id]/+page.svelte`), but it's OK. Just a reminder to be a bit more careful next time! 👍🏻

databases/catdat/data/categories/LRS_R.yaml
ScriptRaccoon: > Regarding the circular arguments: it does make sense that the automated deduction wouldn't create those. I guess what I was thinking there was along the lines of adding annotations to the "soft references" to make them into checkable references. (And the redundancy checker could then also take those into account.) So, something roughly along the lines of:
> 
> ```
> depends:
>   - category: Top
>     property: cartesian filtered colimits
> ```

This is the topic of #301.

PR #237, opened 2026-06-08
Combine content pages about subcategories

ScriptRaccoon:
Three content pages have been combined into one that compiles results about subcategories. Specifically, if $C$ is a suitable subcategory of $D$, and $D$ has a certain property, then $C$ has it, too.


ScriptRaccoon:
FYI @dschepler

Issue #18, opened 2026-03-20
Add more properties of categories


ScriptRaccoon:
As of writing this, CatDat already has 79 properties of categories, but there are many more which are well-known and should be added at some point. IMPORTANT: for each of these, according to the contribution guidelines, one should then also add relevant implications and also decide which categories have this property and which don't. Because of this, this is quite a bit of work. Maybe one can do one commit (or PR) per property.

| Property | Done | 
| -- | -- |
| $\aleph_1$-filtered colimits  | ❌ |
| $\aleph_1$-cofiltered limits  | ❌ |
| pullback-stable coproducts | ❌ |
| total | ❌ |
| cototal | ❌ |
| images | ❌ |
| coimages | ❌ |
| concretizable | ❌ |
| quasi-topos | ❌ |
| has dense small subcategory | ❌ |
| cartesian sifted colimits | ❌ |
| Barr-exact | ✅ |
| gaunt | ✅ |
| regular | ✅ |
| extensive | ✅ |
| generating set | ✅ |
| accessible | ✅ |
| filtered / cofiltered | ✅ |

Further properties which probably are not so relevant for now:

- pre-abelian
- pseudo-abelian
- semi-abelian
- quasi-abelian
- semisimple abelian

varkor:
Another property would be "cartesian sifted colimits". I would like to know some examples of categories that have "cartesian filtered colimits" that do not have "cartesian sifted colimits" (I suspect there aren't many).

varkor:
Perhaps worth noting that the property of having a dense small subcategory is called "bounded" in [Unexpected properties of locally presentable categories](https://link.springer.com/article/10.1007/BF01182450) (not to be confused with [locally bounded categories](https://ncatlab.org/nlab/show/locally+bounded+category)).

PR #243, opened 2026-06-11
Add quasitopos property (jumbo version, to be split out into more focused PRs)

dschepler:
Adds properties: quasitopos, Grothendieck quasitopos
As a prototypical example of the latter, adds the category of separated presheaves on a topological space


databases/catdat/data/category-properties/quasitopos.yaml
```diff
@@ -0,0 +1,16 @@
...
+nlab_link: https://ncatlab.org/nlab/show/quasitopos
+invariant_under_equivalences: true
+
+related_properties:
+  - finitely complete
+  - finitely cocomplete
+  - locally cartesian closed
+  - regular subobject classifier
+  - elementary topos
+  - balanced
```
ScriptRaccoon, line 17: Why "balanced"? This is more "anti-related"?

databases/catdat/data/category-implications/topos.yaml
```diff
@@ -86,3 +86,23 @@
...
+  is_equivalence: true
+
+- id: balanced_quasitopos_is_topos
+  assumptions:
+    - quasitopos
+    - balanced
+  conclusions:
+    - elementary topos
+  proof: This is proven in <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a>, A2.6.3.
+  is_equivalence: true
```
ScriptRaccoon, line 129: We can remove that it is an equivalence, since the other direction follows from the other implications in the db, right?

databases/catdat/data/category-properties/quasitopos.yaml
```diff
@@ -0,0 +1,16 @@
+id: quasitopos
+relation: is a
+description: >-
+  A <i>quasitopos</i> is a category which is finitely complete, finitely cocomplete, and locally cartesian closed, and which has a regular subobject classifier. This gives the category properties similar to those of an elementary topos; a major difference is that a quasitopos need not be balanced.
```
ScriptRaccoon: I think one "and" is too much.

databases/catdat/data/category-properties/quasitopos.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+  Note that some authors use a strong subobject classifier in place of a regular subobject classifier in the definition, i.e. a morphism $\top : 1 \to \Omega$ which classifies strong monomorphisms. For the equivalence of these definitions, see <a href="https://math.stackexchange.com/q/4335533">MSE/4335533</a>.
```
ScriptRaccoon: Please add target=_blank on the link.

ScriptRaccoon: I can also try to write a script that injects this automatically for all external links... WDYT? This is already done for markdown files, but not for YAML files.

databases/catdat/data/category-properties/quasitopos.yaml
ScriptRaccoon, line 1: Can we perhaps also add (directly in this PR) a simple but interesting example of a quasi-topos which is not a topos? Currently the examples are all thin, thus not interesting.

https://catdat.app/category-search/results?satisfied=finitely_complete%7Efinitely_cocomplete%7Elocally_cartesian_closed%7Eregular_subobject_classifier&unsatisfied=elementary_topos%7Ethin

The nlab page mentions some examples.

ScriptRaccoon:
Thank you for the PR! I made some minor comments. I haven't checked out the code yet in my editor, I will do that tomorrow.

---

Related issue: #18

databases/catdat/data/category-properties/quasitopos.yaml
ScriptRaccoon: ok maybe it's OK because you mention this property in the definition text

databases/catdat/data/category-properties/quasitopos.yaml
ScriptRaccoon: I think https://ncatlab.org/nlab/show/subsequential+space would be a nice example.

dschepler:
Actually, could you check Johnstone C2.2.13 for the exact statement, and if it matches what's on the nLab page, could you check whether the definition of "generating set" he uses matches the one we use?  If so, then I'd think the category of pseudotopological spaces would be a counterexample to the first statement: it's certainly a locally small quasitopos; it's cocomplete (for arbitrary coproducts, take the disjoint union and set the convergent filters to be exactly the ones containing an image of a convergent filter in one of the components); and it has a generating set (in fact it's even well-pointed).  I suspect that the correct statement of the first one would have to replace "generating set" with "small dense subcategory".

If that's the case, then we don't currently have all the component properties of either formulation. Though I guess I could approximate it with a combination of the two, "Grothendieck quasitopos <-> quasitopos + locally small + locally presentable".  EDIT: Actually I guess locally presentable -> locally essentially  small so that becomes Grothendieck quasitopos <-> quasitopos + locally presentable.

ScriptRaccoon:
I don't understand the context of your comment. Are you reviewing something which is already in the PR? Is there something wrong? What do you mean with "first statement" in "If so, then I'd think the category of pseudotopological spaces would be a counterexample to the first statement"?

Are you thinking of adding the result C2.2.13 to the implications?

Bildschirmfoto 2026-06-12 um 20 39 46<br />

Here is the definition (urgh...)

Bildschirmfoto 2026-06-12 um 20 46 25

dschepler:
The context was that I was considering adding "Grothendieck quasitopos" property to the PR, and then the example of SepPsh(X) as an example of a Grothendieck quasitopos which is not an elementary topos (under the assumption that the topology on X is not totally ordered under inclusion).  So I wanted to evaluate what would be a correct equivalent condition, hopefully using existing properties.

The "first statement" was referring to the restatement of this theorem on the nLab quasitopos page, which corresponds to the equivalence (ii) <=> (iii) in Johnstone's theorem.

So, it does look like Johnstone's definition of "generating set" is different from ours (which he calls "separating set").  Johnstone's version certainly seems related to being able to express objects as colimits of generating objects, though I'm not sure of the exact relation off the top of my head.

ScriptRaccoon:
I understand. But I won't be able to help with that, I am afraid. Maybe it's also necessary to edit the nLab article or reach out to the authors since it "misquotes" (because of Johnstone's weird terminology) the theorem.

databases/catdat/data/categories/SepPsh(X).yaml
```diff
@@ -0,0 +1,64 @@
...
+  products:
+    description: section-wise defined direct product
+
+special_morphisms:
+  isomorphisms:
+    description: morphisms of separated presheaves that are bijective on every open set
+    proof: This is easy.
+  monomorphisms:
+    description: morphisms of separated presheaves that are injective on every open set
+    proof: ...
```
dschepler: I wonder if Johnstone might have general results that I could refer to here. Along the lines of: in any quasitopos gotten as the $j$-separated objects of a topos for $j$ a Lawvere-Tierney topology, the descriptions in general are:
monomorphisms: same as monomorphisms in the topos
regular monomorphisms: $j$-closed monomorphisms
epimorphisms: morphisms whose image (calculated in the topos) are $j$-dense
regular epimorphisms: same as epimorphisms in the topos

(Or if there's the special case for a Grothendieck quasitopos Bisep(C,J,K) as a corollary somewhere, that would be even better.)

If not, I should be able to prove them myself easily enough on a contents/ page.

ScriptRaccoon: Because of this I have not started to review the new changes. It seems this is still WIP.

databases/catdat/data/category-properties/Grothendieck quasitopos.yaml
```diff
@@ -0,0 +1,10 @@
+id: Grothendieck quasitopos
+relation: is a
+description: Given a small category $\C$ with a pair of Grothendieck topologies $J \subseteq K$, we define $\BiSep(\C, J, K)$ so be the full subcategory of presheaves on $\C$ which are both a sheaf for the $J$ topology and also separated for the $K$ topology. A <i>Grothendieck quasitopos</i> is a category which is equivalent to $\BiSep(\C, J, K)$ for some $\C, J, K$. Equivalently, a category is a Grothendieck quasitopos if and only if it is equivalent to the full subcategory of separated objects for is a Lawvere-Tierney topology on a Grothendieck topos.
```
ScriptRaccoon:
```suggestion
description: Given a small category $\C$ with a pair of Grothendieck topologies $J \subseteq K$, we define $\BiSep(\C, J, K)$ to be the full subcategory of presheaves on $\C$ which are both a sheaf for the $J$ topology and also separated for the $K$ topology. A <i>Grothendieck quasitopos</i> is a category which is equivalent to $\BiSep(\C, J, K)$ for some $\C, J, K$. Equivalently, a category is a Grothendieck quasitopos if and only if it is equivalent to the full subcategory of separated objects for a Lawvere-Tierney topology on a Grothendieck topos.
```

dschepler:
Right now, I think the major property to be resolved is whether it has effective cocongruences. I'm pretty convinced, using that the sheafification is effective, that the comparison morphism $X \sqcup_Y X \to R$ is both monic and epic. So, if I could somehow use the cotransitivity morphism $R \to R \sqcup_{i_2,X,i_1} R$ to show the map $X \sqcup_Y X \to R$ is either regular monic or regular epic, or even that it's strong monic, then that would finish off the proof. I haven't found anything yet along those lines, though.

For the opposite approach, a non-effective equivalence relation I can think of would be something like: say $x \sim y$ for $x,y$ compatible collections of sections over $U_i$ if ($x$ has a gluing if and only if $y$ has a gluing). Or, if that statement in the internal language is satisfied, which is equivalent to: for each $V$, ${x_i}|\_{U_i \cap V}$ has a gluing if and only if ${y_i} \|_{U_i \cap V}$ has a gluing. It seems pretty clear that's not functorial, though: it should be pretty easy to add a gluing of $x$ without adding a gluing of $y$.

But in any case, any counterexample would have to be roughly of this form "equality when restricted to a regular subobject, augmented by existence of something where that something is forced to be unique if it exists".

*Edit: Fixed MathJax formulas*

dschepler:
I think I have a rough proof now: Since $X+X' \to E$ is an epimorphism, that means any section $e \in E(U)$ is locally in one of the copies of $X$. Now $e = r(e)$ is a regular subobject of $y_U$, which must be some $y_V$; and similarly the equalizer of $e = r(e)'$ is some $y_W$. The epimorphism condition implies $U = V \cup W$. We also have that $V \cap W$ is equivalent to $r(e) \in Y$. Now, since $R+R \to R+_X R$ is a regular epimorphism, $t(e)$ must either be in the first or second copy of $R$. If it's in the first copy, then using a splitting of $R \to R +_X R$, we can see it's in fact equal to $i_1(e)$. But then on $W$, we can calculate that $t(e) = i_1(e)$ implies $r(e) = r(e)'$, so $W \subseteq V \cup W$. Thus, $e$ is in the image of $X+X$. And a similar proof holds if $t(e) = i_2(e)$. This shows that $X+X \to E$ is in fact a regular epimorphism; and the category is also regular (any quasitopos is, as the automated deductions suffice to show), so $E$ is effective.

In fact, this should pretty easily generalize to the category of separated objects of a Lawvere-Tierney topology. The only thing that makes it not as easy to generalize to an arbitrary quasitopos is the involvement of regular epimorphisms.

content/separated_objects_special_morphisms.md
ScriptRaccoon, line 1: I think it makes sense to merge the two pages.

content/separated_objects_cocongruences.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Cocongruences in a Quasitopos of Separated Objects
+description: A proof that the full subcategory of separated objects for a Lawvere-Tierney topology on a topos has effective cocongruences
+author: Daniel Schepler
+---
+
+## Cocongruences in a Quasitopos of Separated Objects
+
+::: Lemma
+Let $\T$ be an elementary topos with a Lawvere-Tierney topology $j$. Then in the full subcategory of $j$-separated objects, every cocongruence is effective.
```
ScriptRaccoon: Let's add a link to https://ncatlab.org/nlab/show/Lawvere-Tierney+topology

databases/catdat/data/category-implications/topos.yaml
```diff
@@ -70,12 +70,13 @@
...
+    - thin
+  proof: >-
+    Let $N \coloneqq \coprod_{m \in \IN} 1$ and consider for every $n \in \IN$ the subobject $N_{\geq n} = \coprod_{m \geq n} 1$ of $N$. For $n \leq n'$ we have $N_{\geq n'} \subseteq N_{\geq n}$. There is a (unique, split) epimorphism $N_{\geq n} \to 1$ for every $n$. By assumption, their limit $\lim_n N_{\geq n} \to 1$ is also an epimorphism. But $0 \to \lim_n N_{\geq n}$ is also an epimorphism since any morphism $N_{\geq n} \to X$ factors through one of the coprojections $i_{n'} : 1 \to N_{\geq n}$ and therefore does not extend to a map $N_{\geq n'+1} \to X$. Thus, $0 \to 1$ is an epimorphism. It follows that for any parallel pair $f, g : X \rightrightarrows Y$, the characteristic morphism of the equalizer is equal to $\top \circ {!}_X$ so that $f = g$.
```
ScriptRaccoon: I must say that I do not understand this proof.

1. Can you make more clear where the quasitopos properties are used? (I must admit that this was also a weakness in my proof before for elementary toposes.)
2. Why is $0 \to \lim_n N_{\geq n}$ an epi? Don't we need to take morphisms on the limit for that? Why do you consider $N_{\geq n} \to X$?
3. There seems to be a typo in  $i_{n'} : 1 \to N_{\geq n}$. It should be $n'$.
4. Please write out domain and codomains in $\top \circ {!}_X$. Like $X \to 1 \to \Omega$.

databases/catdat/data/category-properties/Grothendieck quasitopos.yaml
```diff
@@ -0,0 +1,10 @@
+id: Grothendieck quasitopos
+relation: is a
+description: Given a small category $\C$ with a pair of Grothendieck topologies $J \subseteq K$, we define $\BiSep(\C, J, K)$ to be the full subcategory of presheaves on $\C$ which are both a sheaf for the $J$ topology and also separated for the $K$ topology. A <i>Grothendieck quasitopos</i> is a category which is equivalent to $\BiSep(\C, J, K)$ for some $\C, J, K$. Equivalently, a category is a Grothendieck quasitopos if and only if it is equivalent to the full subcategory of separated objects for a Lawvere-Tierney topology on a Grothendieck topos.
```
ScriptRaccoon: Please add a reference for the equivalence.

databases/catdat/data/category-properties/elementary topos.yaml
```diff
@@ -11,3 +11,4 @@ related_properties:
   - natural numbers object
   - subobject classifier
   - pretopos
+  - quasitopos
```
ScriptRaccoon, line 14: We can also add Grothendieck quasitopos

databases/catdat/data/category-implications/topos.yaml
```diff
@@ -86,3 +87,56 @@
...
+  proof: This is proven in <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a>, A2.6.3.
+  is_equivalence: false
+
+- id: Giraud_for_quasitopos
+  assumptions:
+    - Grothendieck quasitopos
+  conclusions:
+    - quasitopos
+    - locally presentable
+  proof: See <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a>, C2.2.13.
```
ScriptRaccoon: Maybe we need a short remark that Johnstone uses a different definition of a generating set, namely a strong generating set, but that any locally presentable category has a strong generating set by Adamek-Rosicky, Theorem 1.20 in Chapter 1?

databases/catdat/data/category-implications/topos.yaml
```diff
@@ -86,3 +87,56 @@
...
+
+- id: Grothendieck_quasitopos_consequences
+  assumptions:
+    - Grothendieck quasitopos
+  conclusions:
+    - exact filtered colimits
+    - cogenerator
+    - effective cocongruences
+  proof: >-
+    In order to calculate colimits in $\BiSep(\C, J, K)$, we can first calculate the colimit in the Grothendieck topos $\Sh(\C, J)$, then take the presheaf quotient of the $K$-closure of the diagonal relation, and finally apply the $J$-sheafification functor. (Note that the last two steps can also be viewed as forming the quotient in $\Sh(\C, J)$, where the equivalence relation is also $J$-closed in a $J$-sheaf and therefore itself a $J$-sheaf.) We already know that filtered colimits in $\Sh(\C, J)$ commute with finite limits, and similarly the $K$-separated quotient functor and $J$-sheafification functor commute with finite limits. We conclude that filtered colimits in $\BiSep(\C, J, K)$ also commute with finite limits.
```
ScriptRaccoon: "In order to calculate colimits [...]" --- Can you add a reference / proof why colimits look like this?

What is the " $K$ closure of the diagonal relation"? And the relation on which object?

Notice that it's the first time for me reading about all these concepts (quasitopos, Grothendieck quasitopos, Lawvere-Tierney topology, etc.), so basically I cannot follow. Can we make the proof accessible to people like me? Can you perhaps add more references in general here?

content/separated_objects_cocongruences.md
ScriptRaccoon:
```suggestion
Let $\T$ be an elementary topos with a Lawvere-Tierney topology $j$. Then in the full subcategory $E$ of $j$-separated objects, every cocongruence is effective.
```

I assume?

databases/catdat/data/categories/SepPsh(X).yaml
```diff
@@ -0,0 +1,70 @@
...
+  - Sh(X)
+
+satisfied_properties:
+  - property: locally small
+    proof: This is easy.
+
+  - property: Grothendieck quasitopos
+    proof: It is equivalent to $\BiSep(\Open(X), J, K)$ where $J$ is the trivial Grothendieck topology and $K$ is the open covering topology.
+
+  - property: cocartesian cofiltered limits
```
ScriptRaccoon, line 24: To clarify this proof, either describe binary coproducts first or even add `property: binary coproducts` with proof before. This will also be useful for `disjoint finite coproducts` below.

databases/catdat/data/categories/SepPsh(X).yaml
```diff
@@ -0,0 +1,70 @@
...
+    proof: For non-empty $U$, both sides of $X \sqcup \lim_{i\in I} Y_i \to \lim_{i\in I} (X \sqcup Y_i)$ can be calculated component-wise. Therefore, for those $U$, the conclusion follows from the corresponding fact in $\Set$. For $U = \varnothing$, we can see that both sides are empty if and only if $X(\varnothing) = \varnothing$ and $Y_i(\varnothing) = \varnothing$ for some $i$, and otherwise both sides are a singleton.
```
ScriptRaccoon: $X$ is already taken. Let us use "sheaf-notation" $F$ and $G_i$.

Also, let's write "both sides ... on $U$".

Replace `$\Set$` by `<a href="/category/Set">$\Set$</a>` since there the proof can be found. Alternatively, link to the implication `extensive_cocartesian_cofiltered_limits`.

databases/catdat/data/categories/SepPsh(X).yaml
```diff
@@ -0,0 +1,70 @@
...
+morphisms: morphisms of presheaves
+description: Here, we assume that the topological space $X$ is such that there is a non-empty family of open subsets whose union is not in the family, since otherwise this category is almost the category of all presheaves. For a few of the properties, we will strengthen this assumption to the assumption that there are two open subsets $U, V$ such that neither is contained in the other.
+nlab_link: https://ncatlab.org/nlab/show/separated+presheaf
+
+tags:
+  - algebraic geometry
+  - topology
+
+related_categories:
+  - Sh(X)
```
ScriptRaccoon, line 14: Let's add **Set**.

databases/catdat/data/categories/SepPsh(X).yaml
ScriptRaccoon, line 1: Do we want to add separate entries for the cases $X = \emptyset$ and $X = *$? It seems like these give categories not already present in the database.

**EDIT.** Ok maybe the case $X = \emptyset$ gives the interval category. For $X = *$ we get **Set** with a formally adjoined initial object, right?

databases/catdat/data/macros.yaml
ScriptRaccoon, line 1: I just noticed that `\Sh: \operatorname{Sh}` is inconsistent. Can you change this to mathbf?

databases/catdat/data/categories/SepPsh(X).yaml
```diff
@@ -0,0 +1,70 @@
...
+
+unsatisfied_properties:
+  - property: skeletal
+    proof: Consider the constant presheaves for two non-equal singleton sets.
+
+  - property: disjoint finite coproducts
+    proof: The equalizer of the two coprojections $1 \rightrightarrows 1 + 1$ has value $1$ at $\varnothing$.
+
+  - property: generator
+    proof: 'The subcategory $\Sh(X)$ of $\SepPsh(X)$ is reflective by <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a> Prop 2.6.12 and A4.4. Therefore, if $\SepPsh(X)$ had a generator then so would $\Sh(X)$, which we know is not the case.'
```
ScriptRaccoon:
```suggestion
    proof: 'The subcategory $\Sh(X)$ of $\SepPsh(X)$ is reflective by <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a>, Prop A.2.6.12 and A.4.4.4. Therefore, if $\SepPsh(X)$ had a generator then so would <a href="/category/Sh(X)">$\Sh(X)$</a>, which we know is not the case.'
```

databases/catdat/data/categories/SepPsh(X).yaml
```diff
@@ -0,0 +1,70 @@
...
+
+  - property: effective congruences
+    proof: 'Let $\{ U_i : i \in I \}$ be a non-empty family of open sets whose union $U$ is not in the family. We then consider the relation $E$ on $X \coloneqq y_U + y_U$ where for $x_1, x_2 \in X(V)$, $(x_1, x_2) \in E(V)$ if and only if either $x_1 = x_2$ or $V \subseteq U_i$ for some $i \in I$. It is easy to see that $E$ is a congruence. However, $E \hookrightarrow X \times X$ is not a regular monomorphism, whereas any effective congruence would necessarily be an equalizer.'
```
ScriptRaccoon: $X$ is already taken.

ScriptRaccoon: Why is $E \to X \times X$ not regular?

databases/catdat/data/categories/SepPsh(X).yaml
```diff
@@ -0,0 +1,70 @@
...
+    proof: Let $U$ and $V$ be two open subsets such that neither is contained in the other. Then there is neither a morphism $y_U \to y_V$ nor a morphism $y_V \to y_U$.
+
+  - property: co-Malcev
+    proof: Let $U$ and $V$ be two open subsets such that neither is contained in the other. We will let $X \coloneqq y_{U \cup V}$, which represents the functor of sections over $U \cup V$, and then consider the reflexive relation on such sections $x, y \in F(U \cup V)$ where $x \sim y$ if and only if there exists $z \in F(U\cup V)$ such that $z |_U = x |_U$ and $z |_V = y |_V$. Note that since $F$ is separated, such a $z$ is unique if it exists. From this, we can see that this relation is representable by the colimit of a diagram where the objects are $y_U$, $y_V$, and three copies of $y_{U\cup V}$, and the morphisms are canonical morphisms from $y_U$ to the first and third copies of $y_{U\cup V}$ and canonical morphisms from $y_V$ to the second and third copies of $y_{U\cup V}$. In fact, we can see that the presheaf colimit is separated, so this presheaf colimit is also the colimit in $\SepPsh(X)$. Thus, we conclude that this corelation is not cosymmetric.
```
ScriptRaccoon: $X$ is already taken

ScriptRaccoon: Why is this relation not cosymmetric?

databases/catdat/data/categories/SepPsh(X).yaml
ScriptRaccoon: ... or refer to the description of coproducts below

dschepler:
The brief introduction to Lawvere-Tierney topologies I would give is: It's just an abstraction of the important properties of the "locally" qualifier on topological spaces. Namely, if $\mathcal{F}' \hookrightarrow \mathcal{F}$ is a sub-presheaf, we say $x \in \mathcal{F}(U)$ is locally in $\mathcal{F}'$ if there exists an open cover $\bigcup_{i\in I} U_i = U$ such that $x|\_{U_i} \in \mathcal{F}'(U_i)$ for each $i$. It is easy to check that the collection of sections satisfying this property is again a sub-presheaf of $\mathcal{F}$, which we might call the "locally"-closure of $\mathcal{F}'$ in $\mathcal{F}$", $cl_{\mathcal{F}}(\mathcal{F'})$. We say $\mathcal{F}'$ is "locally"-closed in $\mathcal{F}$ if that closure is $\mathcal{F}'$, and it is "locally"-dense in $\mathcal{F}$ if that closure is $\mathcal{F}$.

Then the observation which relates this to separatedness and sheaves is: a presheaf $\mathcal{F}$ is separated if and only if for each "locally"-dense inclusion $\mathcal{G}' \hookrightarrow \mathcal{G}$, the restriction function $Hom(\mathcal{G}, \mathcal{F}) \to Hom(\mathcal{G}', \mathcal{F})$ is injective; and $\mathcal{F}$ is a sheaf if and only if for each such "locally"-dense inclusion, the restriction function is bijective. (The forward implication in both cases is a straightforward exercise; the reverse implication can be gotten by considering $\mathcal{G} := y_U$ and $\mathcal{G}' := \bigcup_{i\in I} y_{U_i}$, where a morphism $\mathcal{G}' \to \mathcal{F}$ can be checked to be equivalent to a compatible collection of sections in $\mathcal{F}(U_i)$.)

A commonly used equivalent condition to separatedness is: a presheaf $\mathcal{F}$ is separated if and only if the diagonal subsheaf $\Delta_{\mathcal{F}} \hookrightarrow \mathcal{F} \times \mathcal{F}$ is "locally"-closed. This condition can be informally interpreted as: "if two sections are locally equal, then the sections themselves are equal". A less commonly used equivalent condition to being a sheaf, which you might sometimes see (for example in Toposes, Triples, and Theories) is: $\mathcal{F}$ is a sheaf if and only if it is separated and for each separated presheaf $\mathcal{G}$ which is an extension of $\mathcal{F}$, then $\mathcal{F}$ is "locally"-closed in $\mathcal{G}$.

The "locally" Lawvere-Tierney topology just extends this to subobjects considered as equivalence classes of monomorphisms as usual. It's also straightforward to generalize this to a Grothendieck site, of course. One of the conditions of a Lawvere-Tierney topology is that the closure operation commutes with pullbacks, so it induces an endomorphism of the Sub functor, and thus a morphism $j : \Omega \to \Omega$ satisfying certain conditions.

dschepler:
I was thinking of modifying the proof of the special morphisms to start with a calculation that the image in Sep(j) is the same as the image as calculated in $\mathcal{T}$, whereas the coimage in Sep(j) (i.e. the equalizer of the cokernel pair) is the j-closure of the image as calculated in $\mathcal{T}$. Then, the rest would be automatic from the fact that in a regular category, $f : X \to Y$ is a monomorphism if and only if $X \to im(f)$ is an isomorphism, and $f : X \to Y$ is a regular epimorphism if and only if $im(f) \to Y$ is an isomorphism; and similarly we have the dual result for coregular categories and coimages.

Does this revision sound good to you?

dschepler:
Incidentally, the search http://localhost:5173/category-search/results?satisfied=regular%7Ecoregular&unsatisfied=thin%7Ebalanced also returned the example of **TorsFreeAb** where the description of special morphisms looked very similar, just with saturation in place of "locally"-closure or $j$-closure. I wonder if there might be some more general result that would encompass both cases. I guess I can make the observation that an abelian group is torsion-free if and only if the diagonal is saturated, and the reflector in that case can also be viewed as taking the quotient by the saturation of the diagonal.

databases/catdat/data/categories/SepPsh(X).yaml
```diff
@@ -0,0 +1,77 @@
...
+
+unsatisfied_properties:
+  - property: skeletal
+    proof: Consider the constant presheaves for two non-equal singleton sets.
+
+  - property: disjoint finite coproducts
+    proof: The equalizer of the two coprojections $1 \rightrightarrows 1 + 1$ has value $1$ at $\varnothing$.
+
+  - property: generator
+    proof: 'The subcategory $\Sh(X)$ of $\SepPsh(X)$ is reflective by <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a> Prop A.2.6.12 and A.4.4.4. Therefore, if $\SepPsh(X)$ had a generator then so would <a href="/category/Sh(x)">$\Sh(X)$</a>, which we know is not the case.'
```
ScriptRaccoon:
```suggestion
    proof: 'The subcategory $\Sh(X)$ of $\SepPsh(X)$ is reflective by <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a> Prop A.2.6.12 and A.4.4.4. Therefore, if $\SepPsh(X)$ had a generator then so would <a href="/category/Sh(X)">$\Sh(X)$</a>, which we know is not the case.'
```

ScriptRaccoon: Actually, this motivates me to work on this todo from my long list (I think you have also mentioned something like this previously).

```
  when implication is referenced, add a dependency in the db.
  this is to ensure consistency and prevent delete anomaly.
  I guess we can add a new column to the property_assignments table.
  this can then also be used for automated proofs.
  same with content pages that are referenced.
  basically, every inline reference should be made official.
  ```

**Comment by ScriptRaccoon** [comment]
> The brief introduction to Lawvere-Tierney topologies I would give is: [...]

Thanks a lot!

> I was thinking of modifying the proof of the special morphisms [...]

I don't have an opinion on this right now.

> [...] I wonder if there might be some more general result that would encompass both cases. [...]

Can we perhaps move this to a separate issue or PR?

As for the review, I need to postpone this a few days. I need some pause from this project.

File: databases/catdat/data/category-implications/topos.yaml [hunk 20]
```diff
@@ -86,3 +89,56 @@
...
+
+- id: Grothendieck_quasitopos_consequences
+  assumptions:
+    - Grothendieck quasitopos
+  conclusions:
+    - exact filtered colimits
+    - cogenerator
+    - effective cocongruences
+  proof: >-
+    We have that $\BiSep(\C, J, K)$ is a reflective subcategory of $\Sh(\C, J)$, with the reflector preserving finite limits. (This is a special case of the more general situation of $\Sep(j)$ being a reflective subcategory of $\T$ whenever $j$ is a Lawvere-Tierney topology on an elementary topos $\T$ with the reflector preserving finite limits.) The result follows from Lemma 3 <a href="/content/subcategories">here</a>.
```
ScriptRaccoon: "The result ..." - it is unclear which property you are referring to here since it is an unordered list on the page. So, it is necessary to specify that you are proving that filtered colimits are exact.

ScriptRaccoon: "... general situation ..." - Please add a reference why it is a reflective subcategory, and for the claim that the reflector preserves finite limits.

databases/catdat/data/category-implications/topos.yaml
```diff
@@ -86,3 +89,56 @@
...
+
+    As for having a cogenerator: we have that $\Sh(\C, K)$ is a reflective subcategory of $\BiSep(\C, J, K)$, and the unit of the reflection adjunction is a monomorphism. It follows that the reflector is faithful, and thus that a cogenerator of $\Sh(\C, K)$ is also a cogenerator of $\BiSep(\C, J, K)$.
```
ScriptRaccoon: Please add a reference why it is a reflective subcategory, and why the unit is a mono.

content/separated_objects_quasitopos.md
```diff
@@ -0,0 +1,35 @@
...
+description: Some results concerning the full subcategory of separated objects for a Lawvere-Tierney topology on a topos
+author: Daniel Schepler
+---
+
+## Results Concerning Quasitopoi of Separated Objects
+
+### Special Morphisms
+
+::: Lemma 1
+Let $\T$ be an elementary topos with a <a href="https://ncatlab.org/nlab/show/Lawvere-Tierney+topology" target="_blank">Lawvere-Tierney topology</a> $j$. Then in the full subcategory $\Sep(j)$ of $j$-separated objects:<br>
```
ScriptRaccoon, line 30: Add the definition of a $j$-separated object (before the lemma).

content/separated_objects_quasitopos.md
```diff
@@ -0,0 +1,35 @@
...
+(a) ($\Rightarrow$) This follows from the fact that $\Sep(j)$ is a reflective subcategory of $\T$. ($\Leftarrow$) This is trivial for any subcategory.<br>
+(b) ($\Rightarrow$) Given a morphism $f : X \to Y$, form the image $\im(f)$ in $\T$, which corresponds to a morphism $\chi_{\im(f)} : Y \to \Omega_\T$. Then $j \circ \chi_{\im(f)} \circ f = \top_Y \circ f$ as morphisms $Y \to \Omega_j$, so if $f$ is an epimorphism in $\Sep(j)$, then we conclude $j \circ \chi_{\im(f)} = \top_Y$. ($\Leftarrow$) Given a morphism $f : X \to Y$ of $j$-separated objects whose image in $\T$ is $j$-dense, suppose we have two morphisms $g, h : Y \rightrightarrows Z$ with $g \circ f = h \circ f$. Then since $Z$ is $j$-separated, the equalizer of $g$ and $h$ is $j$-closed; it also contains the image of $f$ and thus is $j$-dense. We conclude that the equalizer is all of $Y$.<br>
```
ScriptRaccoon: Why `$j \circ \chi_{\im(f)} \circ f = \top_Y \circ f$`?

content/separated_objects_quasitopos.md
```diff
@@ -0,0 +1,35 @@
...
+(c) ($\Rightarrow$) Any equalizer in $\T$ of $f, g : X \rightrightarrows Y$ with $Y$ $j$-separated is a $j$-closed subobject of $X$. If $X$ is $j$-separated as well, then that equalizer subobject is automatically separated, and agrees with the equalizer in $\Sep(j)$. ($\Leftarrow$) For a $j$-closed subobject $f : X \hookrightarrow Y$, we see that the characteristic morphism in $\T$, $\chi_X : Y \to \Omega$, factors through $\Omega_j$. Now $X$ is the equalizer of $\chi_X, \top : Y \rightrightarrows \Omega_j$.<br>
```
ScriptRaccoon: `$\chi_X : Y \to \Omega$` - I think it is better to write `\Omega_\T` as before.

content/separated_objects_quasitopos.md
```diff
@@ -0,0 +1,35 @@
...
+
+::: Lemma 1
+Let $\T$ be an elementary topos with a <a href="https://ncatlab.org/nlab/show/Lawvere-Tierney+topology" target="_blank">Lawvere-Tierney topology</a> $j$. Then in the full subcategory $\Sep(j)$ of $j$-separated objects:<br>
+(a) The monomorphisms are the morphisms whose image in $\T$ are monomorphisms.<br>
+(b) The epimorphisms are the morphisms whose image in $\T$ are $j$-dominant (i.e. the image calculated in $\T$ is a $j$-dense subobject of the codomain).<br>
+(c) The regular monomorphisms are the morphisms whose image in $\T$ are $j$-closed monomorphisms.<br>
+(d) The regular epimorphisms are the morphisms whose image in $\T$ are epimorphisms.
+:::
+
+_Proof._ Recall that $\Sep(j)$ is a reflective subcategory of $\T$, where the reflector takes an object $X$ to the quotient of $X$ by the congruence defined by the $j$-closure of the diagonal in $X\times X$. Also recall that the equalizer of $j, \id : \Omega_\T \rightrightarrows \Omega_\T$ is a $j$-separated object $\Omega_j$ which serves as the regular subobject classifier in $\Sep(j)$, and since $j$ is idempotent, this can also be described as the image (in $\T$) of $j$.<br>
```
ScriptRaccoon: Maybe write `X_{sep}` since this notation is used below.

content/separated_objects_quasitopos.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Effective Cocongruences
+
+::: Lemma 2
+Let $\T$ be an elementary topos with a Lawvere-Tierney topology $j$. Then in the full subcategory $\Sep(j)$ of $j$-separated objects, every cocongruence is effective.
+:::
+
+_Proof._ Suppose $p : X + X' \twoheadrightarrow E$ is a cocongruence (where $X'$ is an isomorphic copy of $X$), with coreflexivity morphism $r : E \to X$ and cotransitivity morphism $t : E \to E +_X E'$. (Here $E'$ is an isomorphic copy of $E$, and $E +_X E'$ is the coproduct modulo the relations $p(x') = p(x)'$.) We will also let $Y$ be the subobject of $X$ given by $x : X$ such that $p(x) = p(x')$. Note that $Y$ is $j$-closed in $X$ since $E$ is $j$-separated.
```
ScriptRaccoon: Let's write that $p$ is a congruence in Sep(j) to make the distinction clear.

content/separated_objects_quasitopos.md
```diff
@@ -0,0 +1,35 @@
...
+
+We will first show that $p$ is in fact an epimorphism in $\T$. The argument will be in the internal logic of $\T$. Thus, suppose $e : E$, and let $x \coloneqq r(e) : X$. Then since $p$ is an epimorphism in $\Sep(j)$, we have $j(e = p(x) \lor e = p(x'))$. Also, since $E + E' \to E +_X E'$ is an epimorphism in $\T$, we have $t(e)$ is the image of an element of either $E$ or $E'$. In the first case, applying the section $e \mapsto e, e' \mapsto p(r(e)') : E +_X E' \to E$ of the inclusion $E \to E +_X E'$, and the fact that $t$ is the unique map such that $p(y) \mapsto p(y), p(y') \mapsto p(y')'$ for each $y : X$, we conclude the section composed with $t$ is the identity; thus, we get $t(e) = e$. Now if $e = p(x')$, then from $t(e) = e$ we conclude $x \in Y$, so $e = p(x)$ also. Therefore, $(e = p(x) \lor e = p(x')) \rightarrow e = p(x)$, so $j(e = p(x) \lor e = p(x'))$ implies $j(e = p(x))$, which in turn implies $e = p(x)$ since $E$ is $j$-separated. Similarly, if $t(e)$ is the image of an element of $E'$, then $e = p(x)'$. In either case, we have shown that $e$ is in the image of $p$, as desired.
```
ScriptRaccoon: I must admit that I am not able to review this PR. Too much is unclear to me, especially in this paragraph.

ScriptRaccoon: Several options:

1. We merge this PR either way, I don't need to understand everything.
2. We find another person that checks the details.
3. More details are added to make it self-contained.

dschepler:
I think I've found a proof that a general quasitopos has effective cocongruences. In fact, the core of the argument seemed to generalize to a proof that regular + extensive + quotients of congruences -> effective cocongruences (which doesn't directly apply to SepPsh(X) for example, but the failure to have disjoint coproducts is controlled enough that a cocongruence lives within a subcategory where you can apply that result).

Interestingly, this seems to have made several "not regular" assignments redundant: for LRS_R, Meas, Met_oo, Pos, Prost, Cat, Top. Essentially, I guess the proof provides a way to take a non-effective cocongruence, and construct a pullback diagram where the bottom row is a regular epimorphism but the top row is not. That pullback diagram, more concretely, has the cocongruence $X+X'\to R$ as top row; $X+X'\to E+E'$ as left column; the cotransitivity morphism $t : E\to E+_X E'$ as right column; and $E+E' \to E+_X E'$ as bottom row.

dschepler:
Before this is ready to be merged, there are a couple references that need to be completed. One is to the decomposition of an arbitrary quasitopos into a comma category of a Heyting algebra and a solid quasitopos. The other is to the fact that a solid quasitopos has disjoint finite coproducts and is therefore extensive (though this one is easy enough to prove by hand, if needed). However, I don't have a copy of Johnstone to search for these references.

If you want another reviewer still, after the refactoring of the "quasitopos -> effective cocongruences" implication, then the reviewer I could think of inviting would be Vladimir Sotirov at MSE. He's the author of the proof there that "strong quasitopos <-> regular quasitopos" and has also given answers to some of my questions there that on review are referencing generalizations of my questions to quasitopoi. If you would want him as a reviewer, is there some way we could provide an interim rendering of the PR branch for him to browse, without his needing to set up the repo and pnpm on one of this machines?

ScriptRaccoon:
Ok, I don't even know what a solid quasi-topos is.

Yes let's invite another reviewer.

Netlify offers branch deployments with preview URLs, so a reviewer can check out the URL directly. But this doesn't work with branches in forks. But for the review process I can just copy the branch into the repository itself etc. so this will be doable. (The only issue is that syncing is not so straightforward.)

Just let me know when I should do that.

dschepler:
In brief, for a quasitopos, TFAE: (1) The category has disjoint finite coproducts. (2) The category is extensive. (3) The unique morphism $0 \to 1$ is a regular monomorphism. The term "solid quasitopos" is defined as option (3) probably because it's the easiest to check.

(The proof: (1) <=> (2) is of course because the quasitopos is locally cartesian closed. (1) => (3): $1+1$ being a disjoint coproduct implies $0$ is the equalizer of the two coinclusion morphisms $1 \to 1+1$. (3) => (1): Since $0$ is a strict initial object, we have a pullback square $(0 \to X) \to (0 \to 1)$, so every $!_X : 0 \to X$ is a regular monomorphism. So, since the category is coregular, in the pushout square $(0 \to X) \to (Y \to X+Y)$, the morphisms $X \to X+Y$ and $Y \to X+Y$ are (regular) monomorphisms. That pushout is also a pullback by Johnstone A2.6.3 (or, use the fact that $0 \to 1$ being regular monic implies it's the equalizer of its cokernel pair $1 \rightrightarrows 1+1$, and we have a map $X+Y \to 1+1$). )

Anyway, I have plans for this evening, but hopefully tomorrow I'll be able to sync the PR branch with main and do another cleanup pass.

databases/catdat/data/category-properties/quasitopos.yaml
ScriptRaccoon, line 1: cf. #259, this property now requires at least one tag (listed in `config.yaml` under `category_property_tags`). I think "topos theory" is a good fit.

Same with "Grothendieck quasitopos".

dschepler:
I discovered that my previous supposed proof that a Grothendieck quasitopos has exact filtered colimits was incorrect: in general, the reflector from a topos T to Sep(j) does not preserve equalizers. And in fact, for any topological space $X$ which has a non-compact open subset, then SepPsh(X) does not have exact filtered colimits. That's because if we have a directed set of open subsets whose union $U$ is not in the set, then we can take the directed diagram of parallel pairs $1 \rightrightarrows 1 +_{y_V} 1$ for $V$ in the set. The directed colimit is $1 \rightrightarrows 1 +_{y_U} 1$, with equalizer $y_U$. On the other hand, the equalizers are $y_V$, and the directed colimit of these equalizers is $\bigcup_V y_V$.

I think if every open subset is compact, then SepPsh(X) does have exact filtered colimits.

How do we want to handle putting this observation in the database? It seems more of a specialized assumption than the ones about the topology not being totally ordered by inclusion, or that there are nontrivial open covers, to assume that there is a non-compact open subset (since for instance separated presheaf categories on finite discrete spaces are still interesting).

ScriptRaccoon:
> How do we want to handle putting this observation in the database? [...]

This problem is essentially #7. I don't have a good final answer. But one option is to just specify $X$ to be a very concrete space, say the real numbers. I would say that we don't necessarily need to cover all cases for all parameterized categories, but rather list interesting examples of categories. In particular, I think SepPSh(X) has been added for the purpose of having a quasi-topos which is not a topos, right? For this purpose, it should be fine to specify $X$. And in the proofs, you can decide if you want to make explicit which step uses which property of $X$ (or none at all).

Obviously, I would not say the same for the category of $R$-modules for a ring $R$, since it is used in all sorts of places for general $R$, and here we should cover many cases.

ScriptRaccoon:
Meta comment: consider splitting this PR into smaller chunks, since otherwise merging will be delayed, the review becomes difficult, and rebasing becomes cumbersome.

- add SepPSh(X) (not all properties have to be decided at once)
- add quasi-topos property (only what's necessary to decide the property for many categories)
- add Grothendieck quasi-topos property (only what's necessary to decide the property for many categories)
- fill in the gaps

for example.

dschepler:
OK, if we want to start off with an example of a quasitopos which is neither an elementary topos nor thin, then I could start off with a simpler and more concrete example, such as:
* The category of subsequential spaces.
* The category of pseudotopological spaces. (Simpler definition if you use the formulation in terms of a limit relation between ultrafilters on $X$ and $X$ including that the principal ultrafilter at $x$ has limit $x$, but maybe a bit more complex to show why it's locally cartesian closed.)
* The category of simple (undirected) graphs.
* The category of binary relations (objects are a set $X$ equipped with a binary relation $R \subseteq X$, morphisms are then functions $X \to Y$ respecting the relations).
* The category **Mono** of injective functions of sets, with morphisms being commutative squares.

Any one of these jump out at you as an example you'd like to get in? Personally, I'd lean more towards the graph example as an expansion into a new "graph-theory" tag.

ScriptRaccoon:
Mono > Rel* > Graph would be my personal preference.

*Need a different notation.

databases/catdat/data/category-properties/Grothendieck quasitopos.yaml
```diff
@@ -0,0 +1,14 @@
+id: Grothendieck quasitopos
+relation: is a
+description: Given a small category $\C$ with a pair of Grothendieck topologies $J \subseteq K$, we define $\BiSep(\C, J, K)$ to be the full subcategory of presheaves on $\C$ which are both a sheaf for the $J$ topology and also separated for the $K$ topology. A <i>Grothendieck quasitopos</i> is a category which is equivalent to $\BiSep(\C, J, K)$ for some $\C, J, K$. Equivalently, a category is a Grothendieck quasitopos if and only if it is equivalent to the full subcategory of separated objects for a <a href="https://ncatlab.org/nlab/show/Lawvere-Tierney+topology" target="_blank">Lawvere-Tierney topology</a> on a Grothendieck topos. For the equivalence, see <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a>, C2.2.13.
+nlab_link: https://ncatlab.org/nlab/show/quasitopos
+invariant_under_equivalences: true
+
+related_properties:
```
ScriptRaccoon, line 7: After merging of https://github.com/ScriptRaccoon/CatDat/pull/269:

```suggestion
related:
```

databases/catdat/data/category-properties/Grothendieck quasitopos.yaml
```diff
@@ -0,0 +1,14 @@
+id: Grothendieck quasitopos
+relation: is a
+description: Given a small category $\C$ with a pair of Grothendieck topologies $J \subseteq K$, we define $\BiSep(\C, J, K)$ to be the full subcategory of presheaves on $\C$ which are both a sheaf for the $J$ topology and also separated for the $K$ topology. A <i>Grothendieck quasitopos</i> is a category which is equivalent to $\BiSep(\C, J, K)$ for some $\C, J, K$. Equivalently, a category is a Grothendieck quasitopos if and only if it is equivalent to the full subcategory of separated objects for a <a href="https://ncatlab.org/nlab/show/Lawvere-Tierney+topology" target="_blank">Lawvere-Tierney topology</a> on a Grothendieck topos. For the equivalence, see <a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a>, C2.2.13.
+nlab_link: https://ncatlab.org/nlab/show/quasitopos
```
ScriptRaccoon, line 4:
```suggestion
nlab_link: https://ncatlab.org/nlab/show/quasitopos
dual: null
```

databases/catdat/data/category-properties/quasitopos.yaml
```diff
@@ -0,0 +1,20 @@
+id: quasitopos
+relation: is a
+description: >-
+  A <i>quasitopos</i> is a category which is finitely complete, finitely cocomplete, locally cartesian closed, and has a regular subobject classifier. This gives the category properties similar to those of an elementary topos; a major difference is that a quasitopos need not be balanced.
+
+  Note that some authors use a strong subobject classifier in place of a regular subobject classifier in the definition, i.e. a morphism $\top : 1 \to \Omega$ which classifies strong monomorphisms. For the equivalence of these definitions, see <a href="https://math.stackexchange.com/q/4335533" target="_blank">MSE/4335533</a>.
+nlab_link: https://ncatlab.org/nlab/show/quasitopos
```
ScriptRaccoon, line 7:
```suggestion
nlab_link: https://ncatlab.org/nlab/show/quasitopos
dual: null
```

databases/catdat/data/category-properties/quasitopos.yaml
```diff
@@ -0,0 +1,20 @@
...
+invariant_under_equivalences: true
+
+related_properties:
```
ScriptRaccoon, line 10: After merging of https://github.com/ScriptRaccoon/CatDat/pull/269:

```suggestion
related:
```

dschepler:
Closing this now (but keeping the branch around in my fork for the moment, for reference for the future PRs which will be split out from this one)

PR #250, opened 2026-06-26
Trivial functors

ScriptRaccoon:
This PR adds two trivial functors:

! : Set &rarr; 1
! : Grp &rarr; 1

The second serves as an example of a full functor which is not faithful.


ScriptRaccoon:
I made some typos. I fixed them directly on the main branch and force pushed.

Issue #219, opened 2026-05-29
Discussion: How should CatDat prioritize proof sources?


ScriptRaccoon:
Currently, there are several ways to provide the proof of a property of a category (or a functor):

- (C) directly within CatDat  
- (M) cite [MSE](https://math.stackexchange.com/) or [MO](https://mathoverflow.net/)  
- (P) cite a paper, book, or another form of publication

**Examples.**

- (C): the short proof that [**CRing** is not balanced](https://catdat.app/category/CRing)
- (C): the **very long** proof that [**BOn** has ℵ₁-filtered colimits](https://catdat.app/category/BOn)
- (M): the proof that [**CRing** is not coregular](https://catdat.app/category/CRing)
- (P): the proof that [**Met**<sub>&infin;</sub> is locally ℵ₁-presentable](https://catdat.app/category/Met_oo)

There are even two suboptions of (C), namely a direct proof in the popup or a dedicated content page, but that is a separate issue and not relevant here. (In fact, there used to be four suboptions, including lemmas and PDFs, but those have already been removed.)

My main question is the following:

## Which sources should be preferred? Do we want a fixed preference order?

A natural initial guess would be:

```
P > M > C
```

The reasoning is that publications are generally stable, peer-reviewed, and considered the most reliable form of reference.

However, there is a serious issue with this ordering: accessibility. I am not affiliated with a university, and I assume that many users of CatDat are in a similar situation. This means that access to academic literature is restricted, making it impossible to read the referenced proofs.

Also, in terms of user experience, and in particular speed and ease of verification, (C) is clearly best: the proof is immediately available without any external lookup.

In some cases, proofs are too long to include in (C). In such situations, it is very convenient if there already exists a well-written thread on MSE or MO that can simply be cited. Maybe the preference order also depends on the proof length?

However, there is also a major drawback with citations to MSE, MO, or similar platforms: link rot and content volatility. **Posts can be deleted or edited at any time by their authors**, which means that a reference in CatDat may silently become invalid. This is particularly problematic because such failures are not easily detectable. There is no automated alert that the reference has disappeared or changed.

Taking all of this into account, maybe a better preference order would be:

```
C > P > M
```

But even if we agree on such an ordering, it is still unclear whether we should allow direct copying of proofs from MSE or MO into CatDat.

A related question is therefore:

## Should we disallow one of the mentioned source types?

[...]

PR #254, opened 2026-06-26
Add total/cototal category properties

dschepler:
Unknown categories decided for "total" property:
[category of Z-functors](http://localhost:5173/category/Z)

Unknown categories for "cototal" property:
[category of commutative monoids](http://localhost:5173/category/CMon)
[category of locally ringed spaces](http://localhost:5173/category/LRS_R)
[category of Z-functors](http://localhost:5173/category/Z)


dschepler:
I have some rough ideas on some of the others: on Hausdorff spaces and semigroups, I think I should be able to use an idea similar to the one for Cat to keep control over the images of constant maps.  For CMon, I think the "subdirectly irreducible" property might have to do with limiting the number of maps to it - though I'm not yet at all sure how to translate that into a contradiction.  And on locally ringed spaces, I have a vague idea that I might be able to define a functor whose L(T) would have a number of maps from Spec k which grows faster than possible for any single locally ringed space.

Anyway, no rush on reviewing this - I was just working on this off and on over the past week, and wanted to get the progress so far pushed before resuming work on the quasitopos PR.

databases/catdat/data/categories/CAlg(R).yaml
```diff
@@ -19,7 +19,7 @@ satisfied_properties:
     proof: There is a forgetful functor $\CAlg(R) \to \Set$ and $\Set$ is locally small.
 
   - property: finitary algebraic
-    proof: Take the algebraic theory of a commutative ring.
+    proof: Take the algebraic theory of a commutative $R$-algebra.
```
ScriptRaccoon, line 21: good catch!

databases/catdat/data/category-properties/cototal.yaml
```diff
@@ -0,0 +1,11 @@
+id: cototal
+relation: is
+description: 'A category $\C$ is <i>cototal</i> when its dual category is total, i.e. it is locally essentially small and the contravariant Yoneda embedding $y : \C \to [\C^{\op}, \Set]$ has a left adjoint.'
```
ScriptRaccoon:
```suggestion
description: 'A category $\C$ is <i>cototal</i> when its dual category is total, i.e. it is locally essentially small and the covariant Yoneda embedding $y : \C \to [\C^{\op}, \Set]$ has a left adjoint.'
```

(not sure)

databases/catdat/data/category-properties/total.yaml
```diff
@@ -0,0 +1,11 @@
+id: total
+relation: is
+description: 'A category $\C$ is <i>total</i> when it is locally essentially small and the covariant Yoneda embedding $y : \C^{\op} \to [\C, \Set]$ has a left adjoint. (For a concrete example of how this left adjoint might look, see <a href="/content/Grp_total_explicit_proof">here</a>.)'
```
ScriptRaccoon:
```suggestion
description: 'A category $\C$ is <i>total</i> when it is locally essentially small and the contravariant Yoneda embedding $y : \C^{\op} \to [\C, \Set]$ has a left adjoint. (For a concrete example of how this left adjoint might look, see <a href="/content/Grp_total_explicit_proof">here</a>.)'
```

(not sure)

ScriptRaccoon: cf. https://ncatlab.org/nlab/show/Yoneda+embedding

databases/catdat/data/category-properties/cototal.yaml
dschepler: Actually, the variance is correct on both, it's the notation that's the wrong way around. Whoops...

databases/catdat/data/category-properties/total.yaml
ScriptRaccoon, line 1: cf. #259, this property now requires at least one tag (listed in `config.yaml` under `category_property_tags`). If no existing tags fit, create a new one.

Same with cototal.

dschepler: OK, off the top of my head, totality is closely related to certain non-small colimits and limits existing.  Namely, say a diagram $X : \mathcal{I} \to \mathcal{C}$ (with $\mathcal{I}$ not necessarily small) "has residually small coslices" if for each object $c$ of $\mathcal{C}$, the collection of connected components of $c \downarrow X$ is bijective to a set. Then a category is total if and only if every diagram with residually small coslices has a colimit. And it also has the consequence (but probably not an equivalence) that if $X : \mathcal{I} \to \mathcal{C}$ has the property that the collection of cones $c \to X_i$ is bijective to a small set for each object $c$ of $\mathcal{C}$, then the diagram $X$ has a limit.

Similarly, since the Yoneda embedding always preserves limits, the totality property (at least if you already know the category is complete and cocomplete) is purely about size issues, so something similar to the solution set condition in the Freyd adjoint functor theorem.

So, I'm thinking tags along the lines of "limits", "colimits" and "size".

databases/catdat/data/category-properties/cototal.yaml
```diff
@@ -0,0 +1,11 @@
...
+relation: is
+description: 'A category $\C$ is <i>cototal</i> when its dual category is total, i.e. it is locally essentially small and the contravariant Yoneda embedding $y : \C^{\op} \to [\C, \Set]$ has a left adjoint.'
+nlab_link: https://ncatlab.org/nlab/show/total+category
+dual_property: total
+invariant_under_equivalences: true
+
+related_properties:
+  - cocomplete
+  - complete
+  - locally copresentable
```
ScriptRaccoon, line 22: In case I merge https://github.com/ScriptRaccoon/CatDat/pull/269 first:

```suggestion
id: cototal
relation: is
description: 'A category $\C$ is <i>cototal</i> when its dual category is total, i.e. it is locally essentially small and the contravariant Yoneda embedding $y : \C^{\op} \to [\C, \Set]$ has a left adjoint.'
nlab_link: https://ncatlab.org/nlab/show/total+category
dual: total
invariant_under_equivalences: true

related:
  - cocomplete
  - complete
  - locally copresentable
```

databases/catdat/data/category-properties/total.yaml
```diff
@@ -0,0 +1,11 @@
...
+relation: is
+description: 'A category $\C$ is <i>total</i> when it is locally essentially small and the covariant Yoneda embedding $y : \C \to [\C^{\op}, \Set]$ has a left adjoint. (For a concrete example of how this left adjoint might look, see <a href="/content/Grp_total_explicit_proof">here</a>.)'
+nlab_link: https://ncatlab.org/nlab/show/total+category
+dual_property: cototal
+invariant_under_equivalences: true
+
+related_properties:
+  - cocomplete
+  - complete
+  - locally presentable
```
ScriptRaccoon, line 22: In case I merge https://github.com/ScriptRaccoon/CatDat/pull/269 first:

```suggestion
id: total
relation: is
description: 'A category $\C$ is <i>total</i> when it is locally essentially small and the covariant Yoneda embedding $y : \C \to [\C^{\op}, \Set]$ has a left adjoint. (For a concrete example of how this left adjoint might look, see <a href="/content/Grp_total_explicit_proof">here</a>.)'
nlab_link: https://ncatlab.org/nlab/show/total+category
dual: cototal
invariant_under_equivalences: true

related:
  - cocomplete
  - complete
  - locally presentable
```

ScriptRaccoon:
It would be cool if we can merge this soon. The few remaining cases don't have to be dealt with at this moment.

dschepler:
OK, what does have to be resolved before we can merge it: There are several places where references are missing or incomplete. And the current proof that total -> complete needs to be finished, or replaced with a reference if for some reason finishing off the proof that the construction does give a limit is too complex.

ScriptRaccoon:
> OK, what does have to be resolved before we can merge it: There are several places where references are missing or incomplete. And the current proof that total -> complete needs to be finished, or replaced with a reference if for some reason finishing off the proof that the construction does give a limit is too complex.

Oups I forgot that the proofs are incomplete.

dschepler:
@ScriptRaccoon  I think this PR is ready for review now.

ScriptRaccoon:
> @ScriptRaccoon I think this PR is ready for review now.

Great! I will have a look in the next days.

database/data/category-implications/total.yaml
ScriptRaccoon, line 1: Based on the diff in this file and the content page file in the other commit: clearly, your editor uses different formatting settings than mine. I use VS Code and the Prettier extension; it is also listed in the file `extensions.json`. The formatting settings are defined in the file `.prettierrc`. What is your setup? Can we find a way that we format the files in the same way?

database/data/category-implications/total.yaml
```diff
@@ -2,39 +2,39 @@
...
 # TODO: If adding a property "has a weak terminal object", that can be added to the conclusions here: The identity functor $\C \to \C$ is also a discrete fibration, and its colimit must certainly at least be a weakly terminal object. It's unclear to me at this time whether it must necessarily be a terminal object in case $\C$ is not locally essentially small.
 - id: total_initial
   assumptions:
-  - total
+    - total
   conclusions:
-  - initial object
-  proof: 'The empty diagram is vacuously a discrete fibration; thus, it has a colimit, which must be an initial object.'
+    - initial object
+  proof: The empty diagram is vacuously a discrete fibration; thus, it has a colimit, which must be an initial object.
```
ScriptRaccoon: Notice that the quotes for single line strings are only necessary when the text has some YAML-special characters, typically `:`

This is why I have removed them here.

content/Grp_total_explicit_proof.md
ScriptRaccoon, line 1: I have done quite a bit of rewording here. Please check if this is OK for you. I am afraid that the git diff is useless.

.cspell.json
ScriptRaccoon, line 1: I have the cspell VS code extension which underlines unknown or incorrect words. For unknown / mathematical words I then click `quick fix > add to config`. Can you perhaps also use it?

The extension is useful since it catches lots of typos. (For example, it caught your typo `cocotal` in `content/missing_cogenerator.md` that I fixed in the other commit.)

content/missing_cogenerator.md
```diff
@@ -12,8 +12,10 @@ Let $\C$ be a pointed category with a faithful functor $U: \C \to \Set$. Assume
...
+
+Now assume that $\C$ is locally essentially small and cototal. Using the axiom of choice, we may assume that for each small cardinal $\kappa$, there is at most one element $X \in \F$ such that $\card(U(X)) = \kappa$. Treating $\F$ as a discrete diagram in $\C$, assumption (1) implies that for any object $Y$ of $\C$, the collection of cocones $\F \to Y$ is bijective with a set, since the maps $X \to Y$ with $\card(U(X)) > \card(U(Y))$ must all be zero in such a cocone. Therefore, by G. M. Kelly, <a href="https://www.numdam.org/item/?id=CTGDC_1986__27_2_109_0" target="_blank">A survey of totality for enriched and ordinary categories</a>, Thm. 5.6, $\C$ must have a coproduct $Y$ of all elements of $\F$. But then by assumption (2), there exists $X \in \F$ such that $\card(U(X)) > \card(U(Y))$; and since $\C$ is pointed, the coprojection $X \to Y$ must be split monic and therefore non-zero. Using assumption (1), we get a contradiction. <span class="qed">$\square$</span>
```
ScriptRaccoon, line 21: > Using the axiom of choice, we may assume that for each small cardinal $\kappa$, there is at most one element $X \in \F$ such that $\card(U(X)) = \kappa$. 

Why? If we discard elements from $F$, doesn't this break (2)?

ScriptRaccoon, line 21: Please explain in which way (the dual of) Thm. 5.6 applies. It says cototal => cocompact => hypercocomplete => ..., where does it say that diagrams with "few cocones" have colimits?

database/data/categories/CAlg(R).yaml
```diff
@@ -38,8 +38,8 @@ unsatisfied_properties:
...
+  - property: cototal
+    proof: 'Let $\F$ be the family of commutative $R$-algebras of the form $R \times k$ where $k$ is an infinite field including a quotient field of $R$. Then for any commutative $R$-algebra $A$, we have a distinguished morphism $R \times k \to A$ consisting of the projection to $R$ followed by the unique morphism $R \to A$. Moreover, if we have any morphism $\varphi : R \times k \to A$ which is not equal to the distinguished morphism, that implies that $\varphi(0, 1) \ne 0$, so the rng homomorphism $k \to R \times k \to A$ is injective, implying $\card(U(A)) \ge \card(U(k))$. From here, an argument similar to the one <a href="/content/missing_cogenerator">here</a> gives a contradiction, using the distinguished morphisms in place of zero morphisms.'
```
ScriptRaccoon, line 59: The lemma has been added to unify various proofs. If we only write now "an argument similar to ...", the lemma has lost its purpose. I suggest to either write down a self-contained proof or find a variant of the lemma that handles this. Maybe (a spontaneous guess!) also a reduction to the category of non-unital commutative algebras is possible (which *is* pointed): you are working with augmented algebras here, which are equivalent to non-unital algebras.

database/data/categories/Alg(R).yaml
```diff
@@ -40,8 +40,8 @@ unsatisfied_properties:
   - property: semi-strongly connected
     proof: This is because already the full subcategory <a href="/category/CAlg(R)">$\CAlg(R)$</a> of commutative algebras is not semi-strongly connected.
 
-  - property: cogenerating set
-    proof: 'We apply <a href="/content/missing_cogenerating_sets">this lemma</a> to the collection of $R$-algebras which are fields: If $F$ is an $R$-algebra that is also a field and $A$ is a non-trivial $R$-algebra, any algebra homomorphism $F \to A$ is injective. For every infinite cardinal $\kappa$ the field of rational functions in $\kappa$ variables over some residue field of $R$ has cardinality $\geq \kappa$ and a non-trivial automorphism (swap two variables).'
+  - property: cototal
+    proof: Essentially the same proof as for <a href="/category/CAlg(R)">$\CAlg(R)$</a> works here.
```
ScriptRaccoon, line 45: Try to find a variant of the lemma that takes care of both categories.

database/data/categories/Ring.yaml
```diff
@@ -43,8 +43,8 @@ unsatisfied_properties:
   - property: semi-strongly connected
     proof: This is because already the full subcategory <a href="/category/CRing">$\CRing$</a> does not have this property.
 
-  - property: cogenerating set
-    proof: 'We apply <a href="/content/missing_cogenerating_sets">this lemma</a> to the collection of fields: If $F$ is a field and $R$ is a non-trivial ring, any ring homomorphism $F \to R$ is injective. For every infinite cardinal $\kappa$ the field of rational functions in $\kappa$ variables has cardinality $\geq \kappa$ and a non-trivial automorphism (swap two variables).'
+  - property: cototal
+    proof: 'This is a special case of the proof for <a href="/category/Alg(R)">$\Alg(R)$</a> with $R = \IZ$.'
```
ScriptRaccoon: This is a bit inconvenient for the reader since the proof for Alg(R) also redirects to the proof for CAlg(R). (And, as mentioned, this proof is also not fully self-contained right now.)

database/data/category-properties/total.yaml
ScriptRaccoon, line 1: Is there any source that defines total categories for categories that are not locally small? Kelly works with enriched categories, and $Set$-enriched categories are locally small. The nlab page on total categories only talks about locally small categories, too. It only mentions *In fact, this definition makes sense even when X is not locally small* in section 3.

I find it a bit unfortunate that the current definition in CatDat (1) differs from the one you find in most texts, in particular nlab, and that the left adjoint definition is only mentioned afterwards (2) is quite technical, (3) is presented by two equivalent conditions, both of which look a bit "random" to me.

Do we perhaps want to add "locally essentially small" as a condition?

content/Grp_total_explicit_proof.md
ScriptRaccoon, line 1: Can you please add a bit of context and motivation here in the introduction? Roughly:

1. **Grp** is total by result XYZ
2. More generally, every category that is monadic over **Set** is total by result XYZ
3. Nevertheless, we add a proof for **Grp** here to highlight XYZ and explain XYZ
4. The construction of the left adjoint shows in particular that ...

database/data/categories/Cat.yaml
```diff
@@ -89,6 +86,12 @@ unsatisfied_properties:
       $$\Sub_{\reg}(\{ 0 \to 1 \to 2 \}) \to \Sub_{\reg}(\{ 0 \to 1 \}) \times_{\Sub_{\reg}(\{1\})} \Sub_{\reg}(\{ 1 \to 2 \})$$
       is not injective. Therefore, $\Sub_{\reg} : \Cat^{\op} \to \Set^+$ does not preserve pullbacks, so it cannot be representable.
 
+  - property: cototal
+    proof: >-
+      For each infinite cardinal $\kappa$, choose a simple group $S_\kappa$ of cardinality $\kappa$ (for example the group of permutations of $\kappa$ of finite support which are even). Now consider the ultra-wide pushout diagram $1 \rightrightarrows B S_\kappa$. Then for any small category $\C$, the collection of cocones $1 \rightrightarrows B S_\kappa \to \C$ is bijective with a set: to form any such cocone, we must first choose the object $X$ of $\C$ which is the image of the object of $1$. Then, we must choose the morphisms $S_\kappa \to \End_{\C}(X)$; but for $\kappa > \card(\End_{\C}(X))$, the corresponding morphism must be zero.
```
ScriptRaccoon, line 95: > Now consider the ultra-wide pushout diagram $1 \rightrightarrows B S_\kappa$

Please explain this a bit more. I assume you mean the diagram that consists of the unique morphisms $1 \to B S_\kappa$ for all $\kappa$?

The notation `$1 \rightrightarrows B S_\kappa$` is a bit weird since $\rightrightarrows$ indicates two arrows. Likewise, the notation `$1 \rightrightarrows B S_\kappa \to \C` is a bit weird.

Also, this is not a pushout diagram (it lacks the pushout object).

ScriptRaccoon, line 95: Maybe it's better to denote this group by $A_\kappa$ so that one doesn't think all the time of the symmetric group (which is literally denoted $S_\kappa$ in the literature).

database/data/categories/Cat.yaml
```diff
@@ -89,6 +86,12 @@ unsatisfied_properties:
...
+
+      On the other hand, we claim that $1 \rightrightarrows B S_\kappa$ does not have a pushout in $\Cat$; by G. M. Kelly, <a href="https://www.numdam.org/item/?id=CTGDC_1986__27_2_109_0" target="_blank">A survey of totality for enriched and ordinary categories</a>, Thm. 5.6, this will imply that $\Cat$ is not cototal. To see this, suppose we have a pushout $\C$ of $1 \rightrightarrows B S_\kappa$, and choose a cardinal $\lambda > \card(\Mor(\C))$. Then the coprojection $i_\lambda : B S_\lambda \to \C$ must be split monic, since we can construct a cocone $1 \rightrightarrows B S_\kappa \to B S_\lambda$ in which $B S_\kappa \to B S_\lambda$ corresponds to the zero map for $\kappa \ne \lambda$, and in which $B S_\lambda \to B S_\lambda$ is the identity. It follows that if $X$ is the image in $\C$ of the object of $B S_\lambda$ under $i_\lambda$, then $i_\lambda$ induces an injective map $S_\lambda \to \End_{\C}(X)$. This gives a contradiction since $\lambda > \card(\End_{\C}(X))$ and $S_\lambda$ is a simple group.
```
ScriptRaccoon, line 97: As in my other comment, please explain in which way Thm. 5.6. gives us that the pushout would exist.

ScriptRaccoon, line 97: > `$\lambda > \card(\Mor(\C))$`

Maybe one can briefly mention that this is possible since $C$ is small.

database/data/categories/Cat.yaml
```diff
@@ -89,6 +86,12 @@ unsatisfied_properties:
       $$\Sub_{\reg}(\{ 0 \to 1 \to 2 \}) \to \Sub_{\reg}(\{ 0 \to 1 \}) \times_{\Sub_{\reg}(\{1\})} \Sub_{\reg}(\{ 1 \to 2 \})$$
       is not injective. Therefore, $\Sub_{\reg} : \Cat^{\op} \to \Set^+$ does not preserve pullbacks, so it cannot be representable.
 
+  - property: cototal
+    proof: >-
```
ScriptRaccoon, line 94: I think we can make the proof more conceptual by doing something like: if Cat is cototal, then also 1 / Cat is cototal by result XYZ, and then (adjunction?) also Grp is cototal, which is not true. In fact, you are basically repeating the proof that Grp is not cototal here.

Alternatively, can we use the lemma?

database/data/categories/Cat.yaml
ScriptRaccoon, line 95: Do we maybe want to write Aut(X) instead of End(X)?

database/data/categories/SemiGrp.yaml
```diff
@@ -71,6 +63,14 @@ unsatisfied_properties:
       $$\begin{align*} X & \coloneqq \langle p \mid p^2 = p \rangle,\\ E & \coloneqq \langle p, q \mid p^2 = p,\, q^2 = q,\, pq = q,\, qp = p \rangle, \end{align*}$$
       whose underlying sets are $\{p\}$ and $\{p,q\}$, respectively. Then $X$ represents the functor sending a semigroup $A$ to its idempotents, and $E$ represents the relation on idempotents $a, b$ of $A$ that $ab = b$, $ba = a$. It is easy to check that this defines an equivalence relation (see <a href="https://mathoverflow.net/a/510809" target="_blank">MO/510744</a> for details). Since $p \ne q$ in $E$, the equalizer of the two maps $X \rightrightarrows E$ is the empty semigroup. Therefore, if $E$ were effective, it would be isomorphic to the coproduct $X \sqcup X$, whose underlying set consists of non-empty words in $p,q$ with $p,q$ strictly alternating. In particular, in this coproduct, $pq \ne q$.
 
+  - property: cototal
+    proof: >-
```
ScriptRaccoon, line 70: Most of my comments from  `Cat.yaml` also apply here.

database/data/categories/SemiGrp.yaml
```diff
@@ -71,6 +63,14 @@ unsatisfied_properties:
...
+      The proof is similar to the proof for <a href="/category/Cat">$\Cat$</a>. For each infinite cardinal $\kappa$, let $S_\kappa$ be a simple group of cardinality $\kappa$ (such as the alternating group on $\kappa$). We can then form the ultra-wide pushout diagram $1 \rightrightarrows S_\kappa$ in $\SemiGrp$. For every semigroup $A$, the collection of cocones $1 \rightrightarrows S_\kappa \to A$ is bijective to a set: for every such cocone, we must first choose an idempotent $e$ of $A$ corresponding to the map $1 \to A$. Then, whenever $\kappa > \card(U(A))$, then for $f_\kappa : S_\kappa \to A$ in the cocone, we see
```
ScriptRaccoon, line 71: Explain briefly that $U(-)$ denotes the underlying set.

database/data/categories/SemiGrp.yaml
```diff
@@ -71,6 +63,14 @@ unsatisfied_properties:
...
+      $$N \coloneqq \{g \in S_\kappa : f_\kappa(g) = e\}$$
+      is a normal subgroup of $G$. It must be non-trivial since otherwise $f_\kappa$ would induce an injective group homomorphism from $G$ to a group contained in $A$. Therefore, $N$ is all of $G$, so $f_\kappa$ is the constant map with image $a$.
```
ScriptRaccoon:
```suggestion
      is a normal subgroup of $G$. It must be non-trivial since otherwise $f_\kappa$ would induce an injective group homomorphism from $G$ to a group contained in $A$. Therefore, $N$ is all of $G$, so $f_\kappa$ is the constant map with image $e$.
```

database/data/categories/Cat.yaml
ScriptRaccoon: Maybe one can use a notation like `$X \mathrel{\substack{\rightarrow\\[-0.6ex]\cdots\\[-0.6ex]\rightarrow}} Y$`, a variant of the `\rightrightrightarrows` macro that I have added recently.

database/data/categories/SemiGrp.yaml
```diff
@@ -71,6 +63,14 @@ unsatisfied_properties:
...
+
+      We now claim that $1 \rightrightarrows S_\kappa$ does not have a pushout in $\SemiGrp$; by G. M. Kelly, <a href="https://www.numdam.org/item/?id=CTGDC_1986__27_2_109_0" target="_blank">A survey of totality for enriched and ordinary categories</a>, Thm. 5.6, this will imply that $\SemiGrp$ is not cototal. To see this, suppose we had a pushout $A$, and let $\lambda$ be a cardinal strictly greater than $\card(U(A))$. Then the coprojection $S_\lambda\to A$ must be split monic, since we can construct a cocone $1 \rightrightarrows S_\kappa \to S_\lambda$ such that the map $S_\kappa \to S_\lambda$ is the constant map with image 1 if $\kappa \ne \lambda$, while the map $S_\lambda \to S_\lambda$ is the identity.  But this contradicts the choice of $\lambda$.
```
ScriptRaccoon, line 75: Such a pushout would yield a coproduct of the $S_\kappa$ in **Grp** (or **Mon**, doesn't really matter), which does not exist (by the argument you mention, but in this category it is clearer to write down).

ScriptRaccoon, line 75: "is the constant map with image 1" <-- in the proof for **Cat**, this was called the zero map. Let's make it consistent. I would call it the trivial homomorphism.

database/data/categories/Cat.yaml
ScriptRaccoon, line 95: > `for example the group of permutations of $\kappa$ of finite support which are even`

Maybe formulate it just like in the proof for **SemiGrp**:

`such as the alternating group on $\kappa$`

I have also used the term "alternating group" (on infinite sets) in other proofs. Since the sign only makes sense for permutations with finite support, this condition should be clear.

database/data/categories/Haus.yaml
```diff
@@ -85,6 +84,13 @@ unsatisfied_properties:
...
...Abbildungen in Y konstant</i>. Math. Z. 90 (1965): 152-154. <a href="http://eudml.org/doc/170472" target="_blank">EUMDL</a>), there is a regular Hausdorff space $X_\kappa$ with at least two points such that every continuous map $X_\kappa \to Q_\kappa$ is constant. (The author only states that $X_\kappa$ is regular, but actually, $X_\kappa$ is regular and $T_1$, hence Hausdorff.) Choose a base point $x_\kappa \in X_\kappa$ for each $\kappa$. We can form an ultra-wide pushout diagram $1 \rightrightarrows X_\kappa$ where each morphism $1 \to X_\kappa$ corresponds to $x_\kappa$. Then for any Hausdorff space $Y$, the collection of cocones $1 \rightrightarrows X_\kappa \to Y$ is bijective to a set: if $Y$ is empty, then the collection of cocones is obviously empty. Otherwise, in order to form a cocone, we must first choose $y \in Y$ corresponding to the morphism $1 \to Y$. Then for each $\kappa \ge \card(U(Y))$, $Y$ is homeomorphic to one of the spaces in the product forming $Q_\kappa$. Therefore, there is a morphism $Y \to Q_\kappa$ splitting the projection map $Q_\kappa \to Y$. It follows that the map $X_\kappa \to Y$ is constant, and in fact it must be the constant map with image $y$.
```
ScriptRaccoon, line 94: The citation looks as if the paper proves something about this specific space $Q_k$. Let's make this a bit more clear that this is a result about any T_1 space.

ScriptRaccoon, line 94: > `We can form an ultra-wide pushout diagram $1 \rightrightarrows X_\kappa$ `

See my remarks on `Cat.yaml`

database/data/categories/Haus.yaml
```diff
@@ -85,6 +84,13 @@ unsatisfied_properties:
...
+
+      On the other hand, we claim that $1 \rightrightarrows X_\kappa$ does not have a pushout in $\Haus$; by G. M. Kelly, <a href="https://www.numdam.org/item/?id=CTGDC_1986__27_2_109_0" target="_blank">A survey of totality for enriched and ordinary categories</a>, Thm. 5.6, this will imply that $\Cat$ is not cototal. To see this, suppose we had a pushout $Y$, and let $\lambda \coloneqq \card(U(Y))$. Then the coprojection $X_\lambda \to Y$ is split monic, since we can construct a cocone $1 \rightrightarrows X_\kappa \to X_\lambda$ where the map $X_\kappa \to X_\lambda$ is the constant map with image $x_\lambda$ when $\kappa \ne \lambda$, and the map $X_\lambda \to X_\lambda$ is the identity. But similarly to the previous paragraph, we can show any morphism $X_\lambda \to Y$ must be constant, giving a contradiction since $X_\lambda$ has at least two points.
```
ScriptRaccoon, line 96: > this will imply that $\Cat$ is not cototal

typo

ScriptRaccoon (changes requested):

database/data/category-implications/total.yaml
dschepler: I've been using Qt Creator.

ScriptRaccoon: Can you create a configuration file that your editor understands?

ScriptRaccoon: Or maybe since this is already a dev dependency, does running the command `pnpm format` work for you?

dschepler: Yes, pnpm format seemed to work. On an old checkout of this branch before your changes, it made the formatting changes you indicated to content/Grp_total_explicit_proof.md and database/data/category-implications/total.yaml . (And it also removed an extraneous comma from src/routes/+layout.svelte by the way.)

content/missing_cogenerator.md
dschepler: I don't see why it would, as long as you still keep one of each cardinality.

dschepler: Isn't hypercocomplete exactly that?

database/data/category-properties/total.yaml
dschepler: I thought Kelly's article had some discussion of local smallness issues. and that the main definition there was that for any functor $F : C^{op} \to Set$ (or replace **Set** with the category C is enriched over) there is a colimit of the identity weighted by $F$ (but I didn't want to get into weighted colimits and the formulation in terms of discrete fibrations is pretty much equivalent). But I wasn't at all sure of this.

If you prefer to require locally essentially small, then I would be fine with that, and it would simplify things considerably. In fact, I had that requirement in originally but then changed it because of what I thought I saw in Kelly's article. (It would also disqualify the category of Z-functors, obviously.)

content/Grp_total_explicit_proof.md
dschepler: The context was simply that I was having trouble wrapping my head around the definition just from reading the nLab page without this explicit example as a rough guide.

database/data/categories/Cat.yaml
dschepler: Should be pretty much equivalent given the group of units functor is a coreflector from monoids to groups, right?

ScriptRaccoon: Yes it is equivalent, I just wanted to ask if Aut(X) is better because it is used by default in the context of group actions.

content/missing_cogenerator.md
ScriptRaccoon: Of course you are right.

database/data/category-properties/total.yaml
ScriptRaccoon: Yes, Kelly discusses some smallness issues in the first pages, but he never suggests a definition of a total category that works for categories that are not locally small, right? Did I miss something?

There is another issue: the equivalence between the two (or three, actually) defining conditions of a total category are only proven for locally small categories by Kelly. Maybe the argument works in general, I just don't know. (I haven't digested the proof.) But in any case, we cannot cite Kelly's paper then.

I have gathered all the issues and asked about them now on mathoverflow: https://mathoverflow.net/questions/3278/
EDIT: I mean https://mathoverflow.net/questions/513986/

content/missing_cogenerator.md
ScriptRaccoon: Yes, but it would be nice to recall this briefly (for example, since hypercocomplete is not a widely known term).

content/Grp_total_explicit_proof.md
ScriptRaccoon: Maybe we can then write at least 1-2 sentences like "The definition of a total category is very abstract. We apply it to a specific category like the the category of groups to get a better understanding of it."

database/data/category-properties/total.yaml
dschepler: I think you mean https://mathoverflow.net/questions/513986/ right?

ScriptRaccoon: Yes, sorry

content/Grp_total_explicit_proof.md
dschepler: I guess another alternative would be to post that as an answer to https://mathoverflow.net/questions/98857/two-questions-about-total-categories (which I just noticed happens to be a question you asked a while ago), and then replace the references to a reference to that answer. (Though the question there is about what the left adjoint looks like for **Ring** so I'd have to add some comments that of course you adapt the construction using the canonical co-ring structure on $\mathbb{Z}[x]$ which represents the identity functor Ring -> Ring.)

ScriptRaccoon:
Since #319 is now merged, can you also add the proof references that have been added here? For example, the proof for **SemiGrp** refers to the proof for **Cat**.

database/data/categories/CRing.yaml
ScriptRaccoon, line 1: This will not be necessary anymore after https://github.com/ScriptRaccoon/CatDat/pull/323 is done.

database/data/categories/Ring.yaml
ScriptRaccoon: This will not be necessary anymore after https://github.com/ScriptRaccoon/CatDat/pull/323 is done.

database/data/category-properties/total.yaml
dschepler: Since nobody's chimed in on MO to argue in favor of allowing for some non locally small categories to be total, I think I'll go ahead and revert to requiring it as a precondition.

content/Grp_total_explicit_proof.md
ScriptRaccoon: I think that makes sense. But we don't necessarily have to delete it from CatDat.

Related issue: #219

database/data/category-properties/total.yaml
ScriptRaccoon: I agree.

database/data/category-properties/total.yaml
```diff
@@ -0,0 +1,27 @@
...
+  <ol>
+  <li>The covariant Yoneda embedding
+  $$y : \C \to [\C^{\op}, \Set]$$
+  has a left adjoint. For a concrete example of how such a left adjoint could look, see <a href="/content/Grp_total_explicit_proof">here</a>.
+  <li>Every <a href="https://ncatlab.org/nlab/show/discrete+fibration" target="_blank">discrete fibration</a> $X : \I \to \C$ (with $\I$ not necessarily essentially small) whose fibers are bijective to sets has a colimit in $\C$. (Recall that $X$ is a discrete fibration if for every morphism $f : Y \to X_i$ in $\C$ there exists a unique $\alpha : j \to i$ in $\I$ such that $X_\alpha = f$.)</li>
+  <li>Let $X : \I \to \C$ be a diagram (with $\I$ not necessarily essentially small) such that for each object $Y$ of $\C$, the collection of connected components of the comma category $Y \downarrow X$ is bijective to a set. Then $X$ has a colimit in $\C$.</li>
+  </ol>
+  The equivalence is proven as Thm. 5.2 and 5.5 in G. M. Kelly, <a href="https://www.numdam.org/item/?id=CTGDC_1986__27_2_109_0" target="_blank">A survey of totality for enriched and ordinary categories</a>.
+
+  A general category is called total if it is equivalent to a small category which is total.
```
ScriptRaccoon:
```suggestion
  A general category is called total if it is equivalent to a locally small category which is total.
```

database/data/category-properties/cototal.yaml
```diff
@@ -0,0 +1,27 @@
...
+  <ol>
+  <li>The contravariant Yoneda embedding
+  $$y : \C^{\op} \to [\C, \Set]$$
+  has a left adjoint.
+  <li>Every <a href="https://ncatlab.org/nlab/show/discrete+opfibration" target="_blank">discrete opfibration</a> $X : \I \to \C$ (with $\I$ not necessarily essentially small) whose fibers are bijective to sets has a limit in $\C$. (Recall that $X$ is a discrete opfibration if for every morphism $f : X_i \to Y$ in $\C$ there exists a unique $\alpha : i \to j$ in $\I$ such that $X_\alpha = f$.)</li>
+  <li>Let $X : \I \to \C$ be a diagram (with $\I$ not necessarily essentially small) such that for each object $Y$ of $\C$, the collection of connected components of the comma category $X \downarrow Y$ is bijective to a set. Then $X$ has a limit in $\C$.</li>
+  </ol>
+  The equivalence is proven as Thm. 5.2 and 5.5 in G. M. Kelly, <a href="https://www.numdam.org/item/?id=CTGDC_1986__27_2_109_0" target="_blank">A survey of totality for enriched and ordinary categories</a>.
+
+  A general category is called cototal if it is equivalent to a small category which is cototal.
```
ScriptRaccoon:
```suggestion
  A general category is called cototal if it is equivalent to a locally small category which is cototal.
```

database/data/category-implications/total.yaml
```diff
@@ -0,0 +1,37 @@
+# results on total and cototal categories
+
+- id: total_loc_ess_small
+  assumptions:
+    - total
+  conclusions:
+    - locally essentially small
+  proof: This is trivial.
+  is_equivalence: false
```
ScriptRaccoon, line 9: Since #336 is now merged, this line can be removed. Same for all other `is_equivalence: false`.

content/missing_cogenerator.md
ScriptRaccoon, line 1: Not sure if this will resolve automatically after rebasing, but after #341 proof environments can be written now as:

```md
::: Proof
bla
:::
```

ScriptRaccoon:
@dschepler  Can you give me an update here? Do you plan to continue this PR at some point? If you prefer, I can also try to finish it (based on the comments I gave).

dschepler:
> @dschepler Can you give me an update here? Do you plan to continue this PR at some point? If you prefer, I can also try to finish it (based on the comments I gave).

I've been taking a bit of a break from the project.

Though unfortunately, I had gotten stuck on most of the significant comments that were left to address:
* I wasn't having much luck trying to come up with a good generalization of the lemma on certain large collections of simple objects precluding being total to the case of Ring. (And I couldn't come up with a good argument relating it to Rng, since as far as I recall, the natural reflector is in the wrong direction.)
* I was having trouble trying to put in the clarification that we were using the implication of hypercompleteness in several of the negative proofs, without making it too much of a distraction or too verbose. (Of course, if we implement #308 then that becomes less of an issue; but this PR is already large enough without adding that in. And I guess implementing #308 could also be an eventual clarification to some of the proofs such as Cat and Haus - as far as I know, being total isn't necessarily inherited by slice or coslice categories; but on the other hand, it seems clear that hypercompleteness should be inherited.)

So, even if I returned to this, I'm not sure I could resolve much more than marking the new dependencies.  If you had clearer ideas on what you wanted to see as resolutions for those comments, feel free either to finish off this PR yourself as you suggested, or to post clarifications and I can try to get back to this and finish it off.

ScriptRaccoon:
I see. Please don't let the wish to make everything perfect block you or hold up this PR. If my comments have given you that impression, I'm sorry. They are always meant merely as suggestions for directions we could take, not as requirements for this PR. If something is not feasible right now, that's perfectly fine. A TODO comment in the code is enough to flag it for later.

Specifically, for the first bullet point, I don't think we necessarily need a general lemma. As long as we understand how the proofs work in each relevant example, that is sufficient for now. A general lemma would be *nice to have*, but it is by no means necessary to finish this PR.

As for the second bullet point, the proofs can become distracting or verbose, as you say, if we leave TODO comments to remind ourselves to refactor them once the related properties are added. But if we already have the proofs, let's just write them down. They don't have to be perfect. (And honestly, no proof is ever perfect.)

database/data/categories/Cat.yaml
dschepler: I'm thinking of this as applying the general fact that a morphism $BM \to \mathcal{C}$ is equivalent to a choice of an object $X$ of $\mathcal{C}$ along with a monoid homomorphism $M \to \mathrm{End}_{\mathcal{C}}(X)$.

dschepler:
All right, I think I've taken care of most issues as far as I can, except that I'm still working on making the CAlg(R) proof self-contained instead of expecting the reader to "apply the patches to the lemma".

content/Grp_total_explicit_proof.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Explicit Proof that the Category of Groups is Total
+description: An explicit construction of the left adjoint to the covariant Yoneda embedding on the category of groups
+---
+
+## Explicit Proof that the Category of Groups is Total
```
ScriptRaccoon:
```suggestion
# Explicit Proof that the Category of Groups is Total
```

once you rebase on `main` because of #348

ScriptRaccoon:
> All right, I think I've taken care of most issues 

Great!

> except that I'm still working on making the CAlg(R) proof self-contained

Let me know when you think this is done (or if it should be postponed, which is also OK, just leave a TODO comment).

Then I will have another look at the PR and merge it.

dschepler:
That last step should be done now. It should hopefully be ready for review, and then it will just need a rebase (probably by hand) before it can be merged.

ScriptRaccoon (approved):

ScriptRaccoon:
Alright. I have some more minor adjustments and fixes. Please have a look. (Unfortunately the diff on GitHub is not very helpful.)

Can you try to rebase? I think we should only have a handful of commits in the end. Such as:

- add properties total and cototal (including Grp_total_explicit_proof.md)
- results on total categories
- decide total and cototal for various categories

database/data/categories/Cat.yaml
ScriptRaccoon: I will resolve since it has no high prio and the proof works as is. (I would write it differently.)

database/data/categories/Alg(R).yaml
ScriptRaccoon: I will close here since it has no high prio.

database/data/categories/CAlg(R).yaml
ScriptRaccoon: I will close this since it has no high prio and the proof works as is.

dschepler:
> Alright. I have some more minor adjustments and fixes. Please have a look. (Unfortunately the diff on GitHub is not very helpful.)

Looked good to me.

> Can you try to rebase? I think we should only have a handful of commits in the end. Such as: ...

That's done now.

ScriptRaccoon:
Great! Thanks a lot for your work on this PR.

PR #258, opened 2026-06-28
Add various properties of the category of Banach spaces

### LINKED ISSUE #257: Is this a correct proof that Ban is counital?
I just happened to notice that "counital" is currently not settled for the category of Banach spaces with linear contractions. After a bit of thought, I came up with this possible proof, which I'm not confident in though:
Suppose we have an epimorphism $A \twoheadrightarrow B$ with maps $f : A \to X \sqcup Y$ and $g : B \to X \times Y$ making a commutative square. Then the function $B \to \mathbb{R}$, $b \mapsto |b|_B - |p_1(b)|_X - |p_2(b)|_Y$, is a continuous function, and it's non-negative on the image of $A$. Since the image of $A$ is dense in $B$, that implies the function is non-negative everywhere on $B$, implying that $g$ factors through $X \sqcup Y$. That establishes that $X \sqcup Y \to X \times Y$ is a strong monomorphism.

And assuming that's correct, then that would give lots of examples of strong monomorphisms which are not regular, implying that Ban is not coregular.

ScriptRaccoon:
This PR adds

- the easy proof that **Ban** is not counital (thus, resolves #257)
- the classification of regular epimorphisms in **Ban**
- the proof that **Ban** is regular
- the proof that **Ban** is coregular
- the proof that **Ban** has no regular quotient object classifier
- the proof that **Ban** has effective congruences, hence is Barr-exact

Thanks to @dschepler !

### DIFF
--- a/databases/catdat/data/categories/Ban.yaml
+++ b/databases/catdat/data/categories/Ban.yaml
@@ -51,6 +51,9 @@ unsatisfied_properties:
   - property: unital
     proof: See <a href="https://math.stackexchange.com/questions/5033161" target="_blank">MSE/5033161</a>.
 
+  - property: counital
+    proof: The canonical morphism $X \sqcup Y \to X \times Y$ is surjective, and therefore an epimorphism. Hence, if it were also a strong monomorphism, it would be an isomorphism. However, for example, when $X = Y = \IC$, the norms do not agree.
+
   - property: co-Malcev
     proof: See the comments to <a href="https://mathoverflow.net/a/509555/2841" target="_blank">MO/509552</a>.
 
--- a/databases/catdat/data/categories/Ban.yaml
+++ b/databases/catdat/data/categories/Ban.yaml
@@ -35,6 +35,13 @@ satisfied_properties:
   - property: cocartesian cofiltered limits
     proof: 'If $X$ is a Banach space and $(Y_i)$ is a cofiltered diagram of Banach spaces, the canonical map $X \oplus \lim_i Y_i \to \lim_i (X \oplus Y_i)$ is an isomorphism: Since the forgetful functor $\Ban \to \Vect$ preserves finite coproducts and all limits, and $\Vect$ has the claimed property (see <a href="/category-implication/biproducts_cartesian_filtered_colimits">here</a>), the canonical map is bijective. It remains to show that it is isometric. For $(x,y) \in X \oplus \lim_i Y_i$ the norm in the domain is $|x| + \sup_i |y_i|$, and the norm in the codomain is $\sup_i (|x| + |y_i|)$, and these clearly agree.'
 
+  - property: regular
+    proof: >-
+      It suffices to prove that regular epimorphisms are stable under pullbacks. We will use their classification below.
+      So let $f : V \to W$ be a regular epimorphism of Banach spaces and let $g : T \to W$ be any morphism. We need to show that the projection $V \times_W T \to T$ is a regular epimorphism.
+      Clearly, it is surjective since $f$ is surjective. Now let $t \in T$ be an element of norm $<1$. Since $g$ is a linear contraction, $g(t)$ has norm $<1$. Since $f$ is a regular epimorphism, there is some $v \in V$ with norm $<1$ and $f(v) = g(t)$.
+      Then $(v,t) \in V \times_W T$ has norm $\max(|v|,|t|) < 1$ and provides a preimage of $t$.
+
 unsatisfied_properties:
   - property: skeletal
     proof: This is trivial.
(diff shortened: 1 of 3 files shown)


dschepler:
Looks good to me.

On a related note, I'm pretty convinced that the regular epimorphisms in **Ban** are exactly the linear contractions which are surjective on open unit balls - or equivalently, surjective maps $f : X \to Y$ such that the norm on $Y$ is equal to $\inf \{ |x| \mid f(x) = y \}$. Which should make it pretty easy to show that regular epimorphisms are stable under pullback, and therefore **Ban** is regular.

ScriptRaccoon:
Thanks! I have added the proofs. So we have filled 3 gaps instead of just one here, great :)

ScriptRaccoon:
Is there a reason why we didn't decide if **Ban** has effective congruences so far? (and thus, is Barr-exact)

This looks correct and doable. I will try to write it down.

databases/catdat/data/categories/Ban.yaml
```diff
@@ -83,3 +93,14 @@ special_morphisms:
   regular monomorphisms:
     description: closed embeddings
     proof: The non-trivial direction follows from the <a href="https://math.stackexchange.com/questions/319867" target="_blank">well-known fact</a> that for every closed subspace of a Banach space its quotient space is again a Banach space.
+  regular epimorphisms:
+    description: >-
+      For a linear contraction $f : V \to W$ the following are equivalent: (1) $f$ is a regular epimorphism. (2) $f$ is surjective and the norm on $W$ is given by $|w| = \inf \{|v| : f(v) = w\}$. (3) $f$ is surjective and $f$ maps the open unit ball of $V$ onto the open unit ball of $W$.
```
dschepler: In (3), the part "$f$ is surjective" is redundant since any element of $W$ is a scalar multiple of an element of the open unit ball.

dschepler:
> Is there a reason why we didn't decide if **Ban** has effective congruences so far? (and thus, is Barr-exact)
> 
> This looks correct and doable. I will try to write it down.

I guess at the vector space level, it's clear a congruence has to be congruence modulo some subspace. The question is: does the norm have to match the max norm? Just from the conditions $|(x,y)| \ge \max(|x|,|y|)$, $|(x,x)| = |x|$, $|(x,y)| = |(y,x)|$, and $|(x,z)| \le \max(|(x,y)|, |(y,z)|)$. (If so, then clearly the 0 slice is a closed subspace, so the congruence is exactly the kernel pair of the quotient map.)

dschepler:
Maybe we can also settle "regular quotient object classifier"?  It looks like that means that $\Psi$ would have to be big enough to be able to have a dense image in any closed subspace, which you should be able to rule out with a cardinality argument, using the fact that any Banach space is sequential.

dschepler:
Just noticed: for the description of products, do you need to restrict to the subspace with finite sup-norm, similar to the restriction needed for the description of coproducts?

databases/catdat/data/categories/Ban.yaml
ScriptRaccoon: Thanks. It's even redundant in (2).

Fixed.

dschepler:
How about this for a proof that **Ban** is coregular: suppose we have a regular mono $i : A \hookrightarrow B$ and $f : A \to X$. Then using the construction of the pushout $B \sqcup_A X$ as a quotient of $B \sqcup X$ by the subspace $\{ (i(a), -f(a)) \mid a \in A \}$ (which it's easy to show is closed in this case), we can conclude that $|\overline{(0, x)}| = \inf_{a\in A} (|i(a)| + |x-f(a)|)$. However, using $a = 0$ (or using the fact that $X \to B \sqcup_A X$ is a morphism), we see $|\overline{(0, x)}| \le |x|$. On the other hand, for each $a$, we have $|i(a)| + |x-f(a)| = |a| + |x-f(a)| \ge |f(a)| + |x-f(a)| \ge |x|$. so $|\overline{(0,x)}| \ge |x|$ also. Therefore, the map $X \to B \sqcup_A X$ is an isometric embedding, and since $X$ is complete its image must be closed.

ScriptRaccoon:
Thank you for all the comments! They have been incorporated, see the recent commits.

I have also added a proof that **Ban** has effective congruences, hence is Barr-exact.

What is quite amazing is that this leaves us with just ONE property to decide! Namely, if cocongruences are effective (equivalently, if **Ban** is Barr-coexact).

ScriptRaccoon:
What troubles me a little bit is that some arguments are very similar to **Vect**, but there is a massive difference: **Ban** is not preadditive. I hope that this (non-property) is not used anywhere without us noticing.

dschepler:
For cocongruences, I think the general argument over at the quasitopos PR will again give $p : X \sqcup X' \to E$ as the pullback of the regular epimorphism $E \sqcup E' \to E \sqcup_X E'$ by the cotransitivity morphism, using a slightly different argument to combine the two individual pullback squares in this case (namely that the combination certainly works at the vector level, and then the norm of the pullback agrees with the norm on $X \sqcup X'$ since the map $X \sqcup X' \to E \sqcup E'$ is isometric). So then, $p$ must be a regular epimorphism; so it's a quotient by a closed subspace, which must be contained in the anti-diagonal, so $E$ is equivalent to the cokernel pair of the corresponding subspace of $X$.

ScriptRaccoon:
I am not sure about the cocongruences. I will merge this PR now. If you want to, you can expand your sketch above in a separate PR.

ScriptRaccoon:
Added after merging: The assignment that **Ban** is not co-Malcev is now redundant, since **Ban** is not counital. (Which is nice since the counterexample was somewhat complicated.)

https://github.com/ScriptRaccoon/CatDat/commit/47f079231048b069bdc040b05869209f9347f21a

PR #266, opened 2026-07-04
Add category Mono of sets with a distinguished subset

dschepler:
The purpose here is primarily to provide a simple example of a quasitopos which is neither an elementary topos nor thin. (Of course, a lot of the manual proofs here will go away once we can add a "Grothendieck quasitopos" property; this category is equivalent to the $\lnot\lnot$-separated presheaves on the walking morphism category, where the $\lnot\lnot$ topology is generated by the single morphism $0 \to 1$ forming a covering sieve of $1$.)

### DIFF
--- a/databases/catdat/data/categories/Mono.yaml
+++ b/databases/catdat/data/categories/Mono.yaml
@@ -38,14 +38,14 @@ satisfied_properties:
     proof: >-
       The component-wise limit gives a pair of a set and a subset, and this pair forms the limit in $\Mono$.
 
-      More concretely, suppose we have a diagram $((X_i, X_i'))_{i \in \I}$, with limit cone $p_i : \lim_{i\in\I} X_i \to X_i$ in <a href="/category/Set">$\Set$</a>. Then the limit in $\Mono$ is $(\lim_{i\in\I} X_i, \bigcap_{i\in\I} p_i^{-1}(X_i'))$. (It is easy to see that in the subset component, we can also restrict $i$ to range over a <a href="https://ncatlab.org/nlab/show/weak+multilimit" target="_blank">weakly terminal</a> set of objects of $I$.)
+      More concretely, suppose we have a diagram $((X_i, X_i'))_{i \in \I}$, with limit cone $p_i : \lim_{i\in\I} X_i \to X_i$ in <a href="/category/Set">$\Set$</a>. Then the limit in $\Mono$ is $(\lim_{i\in\I} X_i, \bigcap_{i\in\I} p_i^{-1}(X_i'))$. (It is easy to see that in the subset component, we can also restrict $i$ to range over a <a href="https://ncatlab.org/nlab/show/weak+multilimit" target="_blank">weakly initial</a> set of objects of $\I$.)
     check_redundancy: false
 
   - property: cocomplete
     proof: >-
       We have that $\Mono$ is a reflective subcategory of $\Set^{\rightarrow}$, with the reflector taking a function $f : X \to Y$ to the pair $(Y, \im(f))$. Therefore, the result follows from the fact that <a href="/category/Set_arrow">$\Set^{\rightarrow}$</a> is cocomplete.
 
-      More concretely, suppose we have a diagram $((X_i, X_i'))_{i \in \I}$, with colimit cone $c_i : X_i \to \colim_{i\in\I} X_i$ in <a href="/category/Set">$\Set$</a>. Then the colimit in $\Mono$ is $(\colim_{i\in\I} X_i, \bigcup_{i\in\I} c_i(X_i'))$. (It is easy to see that in the subset component, we can also restrict $i$ to range over a <a href="https://ncatlab.org/nlab/show/weak+multilimit" target="_blank">weakly initial</a> set of objects of $I$.)
+      More concretely, suppose we have a diagram $((X_i, X_i'))_{i \in \I}$, with colimit cone $c_i : X_i \to \colim_{i\in\I} X_i$ in <a href="/category/Set">$\Set$</a>. Then the colimit in $\Mono$ is $(\colim_{i\in\I} X_i, \bigcup_{i\in\I} c_i(X_i'))$. (It is easy to see that in the subset component, we can also restrict $i$ to range over a <a href="https://ncatlab.org/nlab/show/weak+multilimit" target="_blank">weakly terminal</a> set of objects of $\I$.)
     check_redundancy: false
 
   - property: disjoint coproducts
@@ -142,7 +142,7 @@ special_morphisms:
     description: >-
       morphisms $f : (X, X') \to (Y, Y')$ such that $f$ is an injective function from $X$ to $Y$, and $f^{-1}(Y') = X'$ (in particular, if $f : X \to Y$ is an inclusion map, then this is equivalent to $X' = X \cap Y'$)
     proof: >-
-      To show any regular monomorphism satisfies this condition, suppose we have a parallel pair $g, h : (Y, Y') \to (Z, Z')$. To form the equalizer in $\Mono$, first we form the equalizer $f : X \hookrightarrow Y$ in $\Set$ of the underlying functions $g$ and $h$. Then by the construction of limits in the proof of completeness above, the equalizer in $\Mono$ is $(X, f^{-1}(Y'))$. Thus, we see that $f^{-1}(Y') = X'$.
+      To show any regular monomorphism satisfies this condition, suppose we have a parallel pair $g, h : (Y, Y') \rightrightarrows (Z, Z')$. To form the equalizer in $\Mono$, first we form the equalizer $f : X \hookrightarrow Y$ in $\Set$ of the underlying functions $g$ and $h$. Then by the construction of limits in the proof of completeness above, the equalizer in $\Mono$ is $(X, f^{-1}(Y'))$. Thus, we see that $f^{-1}(Y') = X'$.
 
       For the other direction, suppose we are given a morphism $f : (X, X') \to (Y, Y')$ satisfying this condition. We can then construct an equalizer diagram
       $$\begin{CD}
--- a/databases/catdat/data/categories/Mono.yaml
+++ b/databases/catdat/data/categories/Mono.yaml
@@ -1,0 +1,158 @@
+id: Mono
+name: category of sets with a distinguished subset
+notation: $\Mono$
+objects: >-
+  pairs $(X, X')$ where $X$ is a set and $X' \subseteq X$ is a subset
+morphisms: >-
+  a morphism $(X, X') \to (Y, Y')$ is a function $f : X \to Y$ such that $f(X') \subseteq Y'$
+description: >-
+  This is equivalent to the full subcategory of objects $(X, Y, f)$ of <a href="/category/Set_arrow">$\Set^{\rightarrow}$</a> where $f : X \to Y$ is an injective function, i.e. a monomorphism. This explains our notation.
+nlab_link: https://ncatlab.org/nlab/show/M-category#def
+
+tags:
+  - set theory
+
+related:
+  - Set
+  - Set_arrow
+  - FI
+  - Pos
+  - Prost
+
+satisfied_properties:
+  - property: locally small
+    proof: There is a forgetful functor $\Mono \to \Set$, $(X, X') \mapsto X$, and $\Set$ is locally small.
+
+  - property: semi-strongly connected
+    proof: This is immediate from the fact that $\Mono$ is equivalent to a full subcategory of <a href="/category/Set_arrow">$\Set^{\rightarrow}$</a>, and the latter is semi-strongly connected.
+
+  - property: generator
+    proof: >-
+      The object $(1, 0)$ is a generator. This is because it represents the functor taking an object $(X, X')$ to $X$, and taking a morphism $f : (X, X') \to (Y, Y')$ to the function $f : X \to Y$.
+
+  - property: cogenerator
+    proof: >-
+      Consider the forgetful functor $U : \Mono \to \Set$, $(X, X') \mapsto X$. This has right adjoint $R : \Set \to \Mono$, $X \mapsto (X, X)$. Therefore, by the dual of Lemma 9 <a href="/content/subcategories">here</a>, $R$ preserves cogenerators; and in particular, since $\Set$ has a cogenerator, so does $\Mono$.
+
+  - property: complete
+    proof: >-
+      The component-wise limit gives a pair of a set and a subset, and this pair forms the limit in $\Mono$.
+
+      More concretely, suppose we have a diagram $((X_i, X_i'))_{i \in \I}$, with limit cone $p_i : \lim_{i\in\I} X_i \to X_i$ in <a href="/category/Set">$\Set$</a>. Then the limit in $\Mono$ is $(\lim_{i\in\I} X_i, \bigcap_{i\in\I} p_i^{-1}(X_i'))$. (It is easy to see that in the subset component, we can also restrict $i$ to range over a <a href="https://ncatlab.org/nlab/show/weak+multilimit" target="_blank">weakly terminal</a> set of objects of $I$.)
+    check_redundancy: false
+
+  - property: cocomplete
+    proof: >-
+      We have that $\Mono$ is a reflective subcategory of $\Set^{\rightarrow}$, with the reflector taking a function $f : X \to Y$ to the pair $(Y, \im(f))$. Therefore, the result follows from the fact that <a href="/category/Set_arrow">$\Set^{\rightarrow}$</a> is cocomplete.
+
+      More concretely, suppose we have a diagram $((X_i, X_i'))_{i \in \I}$, with colimit cone $c_i : X_i \to \colim_{i\in\I} X_i$ in <a href="/category/Set">$\Set$</a>. Then the colimit in $\Mono$ is $(\colim_{i\in\I} X_i, \bigcup_{i\in\I} c_i(X_i'))$. (It is easy to see that in the subset component, we can also restrict $i$ to range over a <a href="https://ncatlab.org/nlab/show/weak+multilimit" target="_blank">weakly initial</a> set of objects of $I$.)
+    check_redundancy: false
+
+  - property: disjoint coproducts
+    proof: This follows from the fact that coproducts in $\Mono$ can be computed component-wise, along with the fact that <a href="/category/Set">$\Set$</a> has disjoint coproducts.
+
+  - property: locally cartesian closed
+    proof: >-
+      For any object $(S, S')$ of $\Mono$, we can view objects of the slice $\Mono / (S, S')$ as indexed families $(\bigsqcup_{s\in S} X_s, \bigsqcup_{s\in S'} X'_s)$ where $X'_s \subseteq X_s$ for each $s\in S'$. Given any two such objects $X = (\bigsqcup_{s\in S} X_s, \bigsqcup_{s\in S'} X'_s)$ and $Y = (\bigsqcup_{s\in S} Y_s, \bigsqcup_{s\in S'} Y'_s)$, observe that $\Hom_{\Mono / (S, S')}(X, Y)$ is equivalent to the tuples of functions $f \in \prod_{s\in S} \Hom_{\Set}(X_s, Y_s)$ such that whenever $s \in S'$, then $f_s(X_s ') \subseteq Y_s '$. We claim that $\Mono$ has a relative exponential given by
+      $$[X, Y]_S \coloneqq \left(\bigsqcup_{s\in S} [X_s, Y_s], \bigsqcup_{s\in S'} [X_s, Y_s]' \right)$$
+      where
+      $$[X_s, Y_s]' \coloneqq \{ f \in [X_s, Y_s] : f(X_s ') \subseteq Y_s ' \}.$$
+      To check that this works, suppose we have a test object $U = (\bigsqcup_{s\in S} U_s, \bigsqcup_{s\in S'} U_s ')$. Then $U \times_S X = (\bigsqcup_{s\in S} (U_s \times X_s), \bigsqcup_{s\in S'} (U_s ' \times X_s '))$. Given a morphism $g : U \times_S X \to Y$ in $\Mono / (S, S')$, we define the corresponding morphism $h : U \to [X, Y]_S$ to have $s$-component given by $u \mapsto (x \mapsto g_s(u, x))$. To check this is a valid morphism in $\Mono / (S, S')$, suppose that $s \in S'$ and $u \in U_s '$; then for every $x \in X_s '$, we have $(u, x) \in U_s ' \times X_s '$, so $g_s(u, x) \in Y_s '$. This shows that $h_s(u) \in [X_s, Y_s]'$. Conversely, given a morphism $h : U \to [X, Y]_S$ in $\Mono / (S, S')$, we define the corresponding morphism $g : U \times_S X \to Y$ to have $s$-component given by $(u, x) \mapsto h_s(u)(x)$. To check this is a valid morphism in $\Mono / (S, S')$, suppose that $s \in S'$ and $(u, x) \in U_s ' \times X_s '$; then $u \in U_s '$ so $h_s(u) \in [X_s, Y_s]'$. Furthermore, $x \in X_s '$, so $h_s(u)(x) \in Y_s '$.
+
+      It is easy to check these give inverse operations which are natural in $U$ (in fact, the details are exactly the same as the details of an explicit proof that $\Set$ is locally cartesian closed), completing the proof.
+
+  - property: regular subobject classifier
+    proof: 'The object $(\{ \top, \bot \}, \{ \top, \bot \})$ is a regular subobject classifier, with $\top : (1, 1) \to (\{ \top, \bot \}, \{ \top, \bot \})$ consisting of the function with image $\top$. This follows from the description of regular monomorphisms below.'
+
+  - property: locally finitely presentable
+    proof: >-
+      It is equivalent to the category of models of the finite limit sketch with a single limit cone
+      $$\begin{CD}
+      X @> \id >> X \\
+      @V \id VV @VV f VV \\
+      X @>> f > Y.
+      \end{CD}$$
+      Alternately, we already know $\Mono$ is cocomplete. We claim that both $(1, 0)$ and $(1, 1)$ are finitely presentable objects of $\Mono$. For the first, $(1, 0)$ represents the forgetful functor $(X, X') \mapsto X$; and this functor has a right adjoint $X \mapsto (X, X)$ so it preserves filtered colimits (in fact all colimits). For the second, $(1, 1)$ represents the functor $(X, X') \mapsto X'$. Now by the previous description of colimits, for a filtered diagram $((X_i, X_i'))_{i\in\I}$, first form the filtered colimit of $X_i$ in <a href="/category/Set">$\Set$</a> with cocone $c_i : X_i \to \colim_{i\in\I} X_i$. Then the filtered colimit in $\Mono$ is $(\colim_{i\in\I} X_i, \bigcup_{i\in\I} c_i(X_i'))$. This makes it easy to see that the canonical function
+      $$\colim_{i\in\I} \Hom((1,1), (X_i,X_i')) \to \Hom((1,1), \colim_{i\in\I} (X_i,X_i'))$$
+      is surjective. To see it is injective, use the fact that the unique morphism $(1, 0) \to (1, 1)$ is an epimorphism (see below), and we have already seen that the corresponding function for $(1,0)$ is injective.
+
+      Therefore, any finite coproduct of $(1, 0)$ and $(1, 1)$ is finitely presentable; this means any object of the form $([n], [m])$ with $m, n \in \IN$ and $m \le n$ is finitely presentable. All such objects form a small set. Moreover, any object $(X,X')$ of $\Mono$ is a filtered colimit of the subobjects $(F, F \cap X')$ where $F$ is a finite subset of $X$, and each $(F, F \cap X')$ is isomorphic to some $([n], [m])$.
+
+  - property: coregular
+    proof: >-
+      It suffices to show regular monomorphisms are stable under pushouts. Thus, suppose we have two morphisms $f : (X, X') \to (Y, Y')$ and $g : (X, X') \to (Z, Z')$ where $f$ is a regular monomorphism, which means $f$ is injective and $X' = f^{-1}(Y')$. Then we can construct a pushout diagram
+      $$\begin{CD}
+      (X, X') @> f >> (Y, Y') \\
+      @V g VV @VV i_1 VV \\
+      (Z, Z') @>> i_2 > ((Y \setminus f(X)) \sqcup Z, i_1(Y') \cup i_2(Z')).
+      \end{CD}$$
+      Here $i_1$ sends $f(x) \mapsto g(x) \in Z$ for $x \in X$, and otherwise it sends $y \mapsto y$ for $y \in Y \setminus f(X)$; and $i_2$ is the inclusion of $Z$. We certainly have $i_2$ is injective; it remains to show $i_2^{-1}(i_1(Y') \cup i_2(Z')) = Z'$. However, we have
+      $$i_2^{-1}(i_1(Y') \cup i_2(Z')) = i_2^{-1}(i_1(Y')) \cup i_2^{-1}(i_2(Z'))$$
+      and obviously $i_2^{-1}(i_2(Z')) = Z'$. Therefore, it suffices to show $i_2^{-1}(i_1(Y')) \subseteq Z'$. Thus, suppose we have $z \in Z$ such that $i_2(z) \in i_1(Y')$. That means there exists $y' \in Y'$ such that $z = i_1(y')$. But in order for $i_1(y')$ to land in the $Z$ portion of $(Y \setminus f(X)) \sqcup Z$, we must have $y' \in f(X)$. That means there exists $x \in X$ such that $f(x) = y'$; and by the assumption on $f$, in fact $x \in X'$. Therefore,
+      $$i_2(g(x)) = i_1(f(x)) = i_1(y') = z = i_2(z),$$
+      so $z = g(x) \in Z'$.
+
+  - property: co-Malcev
+    proof: >-
+      Suppose we have a coreflexive corelation $p : (X \sqcup X, X' \sqcup X') \twoheadrightarrow (E, E')$ with coreflexivity morphism $r : (E, E') \to (X, X')$. From the assumption that $p$ is an epimorphism, we have that $p : X \sqcup X \to E$ is a surjective function. Since <a href="/category/Set">$\Set$</a> is co-Malcev, it follows that $E \cong X \sqcup_Y X$ for some subset $Y \subseteq X$. It remains to show that $E' = i_1(X') \cup i_2(X') \subseteq X \sqcup_Y X$. Certainly, since we have a morphism $(X \sqcup_Y X, i_1(X') \cup i_2(X')) \to (E, E')$ induced by $p$, we must have $i_1(X') \cup i_2(X') \subseteq E'$. On the other hand, any element of $E'$ is an element of $E$ and hence is equal to either $i_1(x)$ or $i_2(x)$ for $x \in X$. In the first case, we must have $x = r(i_1(x)) \in X'$, so $i_1(x) \in i_1(X')$; and similarly for the second case.
+
+  - property: effective cocongruences
+    proof: 'See the proof that $\Mono$ is co-Malcev: It shows that in fact any coreflexive corelation is equivalent to an effective cocongruence $X \sqcup X \twoheadrightarrow X \sqcup_Y X$.'
+
+unsatisfied_properties:
+  - property: skeletal
+    proof: Consider the objects $(X, X)$ and $(Y, Y)$ for isomorphic but non-equal sets $X$ and $Y$.
+
+  - property: balanced
+    proof: The unique morphism from $(1, 0)$ to $(1, 1)$ is both a monomorphism and an epimorphism, but not an isomorphism (see descriptions below).
+
+  - property: cofiltered-limit-stable epimorphisms
+    proof: >-
+      We already know that <a href="/category/Set">$\Set$</a> does not have this property. Now apply the contrapositive of the dual of Lemma 2 <a href="/content/subcategories">here</a> to the functor $\Set \to \Mono$ which sends a set $X$ to the pair $(X, X)$.
+
+  - property: Malcev
+    proof: This is clear since $\Set$ is not Malcev, and it has a left exact embedding in $\Mono$ via the functor sending a set $X$ to $(X, X)$.
+
+special_objects:
+  initial object:
+    description: $(0, 0)$
+  terminal object:
+    description: $(1, 1)$
+  coproducts:
+    description: component-wise defined disjoint union
+  products:
+    description: component-wise defined cartesian product
+
+special_morphisms:
+  isomorphisms:
+    description: >-
+      morphisms $f : (X, X') \to (Y, Y')$ such that $f$ is a bijection between $X$ and $Y$, and $f(X') = Y'$
+    proof: This is easy.
+  monomorphisms:
+    description: >-
+      morphisms $f : (X, X') \to (Y, Y')$ such that $f$ is an injective function from $X$ to $Y$
+    proof: >-
+      For the non-trivial direction, use the fact that the functor taking an object $(X, X')$ to $X$ and a morphism $f : (X, X') \to (Y, Y')$ to the function $f : X \to Y$ is representable by $(1, 0)$.
+  epimorphisms:
+    description: >-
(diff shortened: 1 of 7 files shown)


databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+morphisms: >-
+  a morphism $(X, Y, f) \to (X', Y', f')$ is a pair of functions $\ell : X \to X'$ and $r : Y \to Y'$ making a commutative square
+  $$\begin{CD}
+  X @>{f}>> Y \\
+  @V{\ell}VV @VV{r}V \\
+  X' @>>{f'}> Y'
+  \end{CD}$$
+description: >-
+  This is the full subcategory of $\Set^{\rightarrow}$ consisting of the injective set functions. It is also equivalent to the category of pairs of a set $X$ and a subset $S \subseteq X$, with morphisms $(X, S) \to (Y, T)$ being the functions $f : X \to Y$ such that $f(S) \subseteq T$.
+nlab_link: https://ncatlab.org/nlab/show/Mono
```
ScriptRaccoon: The link doesn't seem to fit.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+related_categories:
+  - Set_arrow
+  - FI
+
+satisfied_properties:
+  - property: locally small
+    proof: This is easy.
+
+  - property: semi-strongly connected
+    proof: This is immediate from the fact that $\Mono$ is a full subcategory of $\Set^{\rightarrow}$, and the latter is semi-strongly connected.
```
ScriptRaccoon: Please add a link to the category page. This way, the category and hence the proof can be found without scrolling anywhere else first.

```
<a href="/category/Set_arrow">$\Set^{\rightarrow}$</a>
```

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+
+  - property: generator
+    proof: >-
+      The object $0 \hookrightarrow 1$ is a generator. This is because it represents the codomain functor taking an object $f : X \hookrightarrow Y$ to $Y$, and taking a morphism $(\ell, r)$ to $r$. That functor is faithful because once $r$ is fixed, $\ell$ is uniquely determined since $f'$ is a monomorphism in $\Set$.
+
+  - property: cogenerator
+    proof: >-
+      The object $\{ \top, \bot \} \xhookrightarrow{\id} \{ \top, \bot \}$. To see this, suppose we have two morphisms $(\ell_1, r_1), (\ell_2, r_2) : (X, Y, f) \rightrightarrows (X', Y', g)$ such that for every morphism $(\ell', r') : (X', Y', g) \to (\{ \top, \bot \}, \{ \top, \bot \})$, $(\ell', r') \circ (\ell_1, r_1) = (\ell', r') \circ (\ell_2, r_2)$. Then morphisms $(X', Y', g) \to (\{ \top, \bot \}, \{ \top, \bot \}, \id)$ are equivalent to functions $Y' \to \{ \top, \bot \}$. Since $\{ \top, \bot \}$ is a cogenerator in $\Set$, this implies $r_1 = r_2$; and since $g$ is a monomorphism in $\Set$, we automatically get $\ell_1 = \ell_2$ as well.
```
ScriptRaccoon: The first sentence seems to be incomplete.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+satisfied_properties:
+  - property: locally small
+    proof: This is easy.
+
+  - property: semi-strongly connected
+    proof: This is immediate from the fact that $\Mono$ is a full subcategory of $\Set^{\rightarrow}$, and the latter is semi-strongly connected.
+
+  - property: generator
+    proof: >-
+      The object $0 \hookrightarrow 1$ is a generator. This is because it represents the codomain functor taking an object $f : X \hookrightarrow Y$ to $Y$, and taking a morphism $(\ell, r)$ to $r$. That functor is faithful because once $r$ is fixed, $\ell$ is uniquely determined since $f'$ is a monomorphism in $\Set$.
```
ScriptRaccoon: Please add what f' is. I think one can write (l,r) : f \to f'

databases/catdat/data/categories/Mono.yaml
ScriptRaccoon: Isn't this an adjunction argument? Can we reuse something?

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+  - property: cogenerator
+    proof: >-
+      The object $\{ \top, \bot \} \xhookrightarrow{\id} \{ \top, \bot \}$. To see this, suppose we have two morphisms $(\ell_1, r_1), (\ell_2, r_2) : (X, Y, f) \rightrightarrows (X', Y', g)$ such that for every morphism $(\ell', r') : (X', Y', g) \to (\{ \top, \bot \}, \{ \top, \bot \})$, $(\ell', r') \circ (\ell_1, r_1) = (\ell', r') \circ (\ell_2, r_2)$. Then morphisms $(X', Y', g) \to (\{ \top, \bot \}, \{ \top, \bot \}, \id)$ are equivalent to functions $Y' \to \{ \top, \bot \}$. Since $\{ \top, \bot \}$ is a cogenerator in $\Set$, this implies $r_1 = r_2$; and since $g$ is a monomorphism in $\Set$, we automatically get $\ell_1 = \ell_2$ as well.
+
+  - property: complete
+    proof: The component-wise limit of injective functions is again an injective function, and it forms the limit in $\Mono$.
+    check_redundancy: false
+
+  - property: cocomplete
+    proof: 'We have that $\Mono$ is a reflective subcategory of $\Set^{\rightarrow}$, with the reflector taking a function $f : X \to Y$ to the inclusion map $\im(f) \hookrightarrow Y$. Therefore, the result follows from the fact that $\Set^{\rightarrow}$ is cocomplete.'
```
ScriptRaccoon: Let's add also here a link to the category page (last sentence).

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+
+      We easily see that $\im(X \times_{D'} Y \to Z \times_D W) \subseteq \im(f) \times_{D'} \im(g)$ (this is the inclusion that we automatically get from the canonical comparison morphism). Conversely, suppose we have $z\in \im(f)$ and $w\in \im(g)$ with the same degree (which must be in $D'$). Then there exist $x\in X$ such that $f(x) = z$ and $y\in Y$ such that $g(y) = w$. Since $\deg_X(x) = \deg_Z(z) = \deg_W(w) = \deg_Y(y) \in D'$, we have $(x, y) \in X \times_{D'} Y$, and the image of $(x, y)$ in $Z \times_D W$ is exactly $(z, w)$, finishing the proof of the claim.
+
+      Therefore, from Lemma 5 <a href="/content/subcategories">here</a>, we conclude that each slice of $\Mono$ is cartesian closed, i.e. that $\Mono$ is locally cartesian closed.
+
+  - property: regular subobject classifier
+    proof: 'The object $\{ \top, \bot \} \xhookrightarrow{\id} $\{ \top, \bot \}$ is a regular subobject classifier, with $\top : (1, 1, \id) \to (\{ \top, \bot \}, \{ \top, \bot \}, \id)$ consisting of the maps with image $\top$. This follows from the description of regular monomorphisms below.'
```
ScriptRaccoon: LaTeX issue: a dollar needs to be removed

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+    proof: This is easy.
+  monomorphisms:
+    description: pairs $(\ell, r)$ where $r$ is an injective function, from which it automatically follows that $\ell$ is also injective
+    proof: For the non-trivial direction, use the fact that the codomain functor taking an object $(X, Y, f)$ to $Y$ and a morphism $(\ell, r)$ to $r$ is representable by $0 \hookrightarrow 1$.
+  epimorphisms:
+    description: pairs $(\ell, r)$ where $r$ is a surjective function
+    proof: >-
+      Suppose we have a morphism $(\ell, r) : (X, Y, f) \to (X', Y', g)$ such that $r$ is surjective. To see it is an epimorphism, suppose we have two morphisms $(\ell_1, r_1), (\ell_2, r_2) : (X', Y', g) \rightrightarrows (T, U, h)$ such that $(\ell_1, r_1) \circ (\ell, r) = (\ell_2, r_2) \circ (\ell, r)$. Then $r_1 \circ r = r_2 \circ r$ implies $r_1 = r_2$. It automatically follows that $\ell_1 = \ell_2$ since $h$ is a monomorphism in $\Set$.
+
+      Conversely, suppose we have an epimorphism $(\ell, r) : (X, Y, f) \to (X', Y', g)$. Then for any set $T$, the morphisms $(X', Y', g) \to (T, T, \id_T)$ are equivalent to the functions $Y' \to T$; therefore,  $r$ is an epimorphism in $\Set$.
```
ScriptRaccoon: Isn't this an adjunction argument?

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+  regular monomorphisms:
+    description: >-
+      pairs $(\ell, r)$ where $r$ is an injective function and moreover the commutative square
+      $$\begin{CD}
+      X @> f >> Y \\
+      @V \ell VV @VV r V \\
+      X' @>> g > Y'
+      \end{CD}$$
+      is in fact a cartesian square
+    proof: >-
```
ScriptRaccoon, line 144: I think the proof will be better if we decide to either

1. work with injective functions, as in the definition, or
2. work with the equivalent category of set-subset pairs

The current approach is somewhere in between and I find it a bit confusing, but also inaesthetic.

I think 2. simplifies the proof, and also the formulation of the regular monos. (This also applies to the classification of monos and epis from above.)

On the other hand, from the perspective of ETCS, there is no difference at all between 1. and 2.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+  regular epimorphisms:
+    description: pairs $(\ell, r)$ where both $\ell$ and $r$ are surjective functions
+    proof: >-
+      To show any regular epimorphism satisfies this condition, suppose we have a parallel pair
+      $$\begin{CD}
+      X @> f >> Y \\
+      @V \ell_1 V \ell_2 V @V r_1 V r_2 V \\
+      X' @> g >> Y'.
+      \end{CD}$$
+      To form the coequalizer in $\Mono$, first we may form the coequalizer in $\Set^{\rightarrow}$, $\coeq(\ell_1, \ell_2) \xrightarrow{\bar g} \coeq(r_1, r_2)$. Note this is not injective in general; however, we may apply the reflector to get the coequalizer $\im(\bar g) \hookrightarrow \coeq(r_1, r_2)$ in $\Mono$. Here, clearly the map $Y' \to \coeq(r_1, r_2)$ is surjective. Likewise, both the functions $X' \to \coeq(\ell_1, \ell_2)$ and $coeq(\ell_1, \ell_2) \to \im(\bar g)$ are surjective, so their composition $X' \to \coeq(\ell_1, \ell_2)$ is surjective as well.
```
ScriptRaccoon: `$coeq(\ell_1, \ell_2)`

Here `\` is missing.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+      \end{CD}$$
+      Here $\top_{X'}$ represents the constant function on $X'$ with value $\top$, and similarly for $\top_{Y'}$. Here, the assumption $X = X' \cap Y$ is used to establish the commutativity of the characteristic morphism $(\chi_X, \chi_Y) : (X', Y', g) \to (\{ \top, \bot \}, \{ \top, \bot \}, \id)$.
+  regular epimorphisms:
+    description: pairs $(\ell, r)$ where both $\ell$ and $r$ are surjective functions
+    proof: >-
+      To show any regular epimorphism satisfies this condition, suppose we have a parallel pair
+      $$\begin{CD}
+      X @> f >> Y \\
+      @V \ell_1 V \ell_2 V @V r_1 V r_2 V \\
+      X' @> g >> Y'.
```
ScriptRaccoon:
```suggestion
      X' @>> g > Y'.
```

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+
+unsatisfied_properties:
+  - property: skeletal
+    proof: Consider the objects $X \xhookrightarrow{\id} X$ and $Y \xhookrightarrow{\id} Y$ for isomorphic but non-equal sets $X$ and $Y$.
+
+  - property: balanced
+    proof: The unique morphism from $0 \hookrightarrow 1$ to $1 \hookrightarrow 1$ is both a monomorphism and an epimorphism, but not an isomorphism (see descriptions below).
+
+  - property: cofiltered-limit-stable epimorphisms
+    proof: >-
```
ScriptRaccoon, line 110: Isn't this an adjunction argument? Can we reuse something?

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+  - property: cocomplete
+    proof: 'We have that $\Mono$ is a reflective subcategory of $\Set^{\rightarrow}$, with the reflector taking a function $f : X \to Y$ to the inclusion map $\im(f) \hookrightarrow Y$. Therefore, the result follows from the fact that $\Set^{\rightarrow}$ is cocomplete.'
+    check_redundancy: false
+
+  - property: disjoint coproducts
+    proof: This follows from the fact that coproducts in $\Mono$ can be computed component-wise, along with the fact that $\Set$ has disjoint coproducts.
+
+  - property: locally cartesian closed
+    proof: >-
+      For this construction, we will work in the equivalent category of pairs $(X, X')$ where $X' \subseteq X$. Now, for any object $(D, D')$ of $\Mono$, the reflector functor $\Set^{\rightarrow} \to \Mono$ induces a reflector functor of slices $\Set^{\rightarrow} / (D' \hookrightarrow D)$ to $\Mono / (D, D')$. We claim that this reflector preserves binary products, i.e. it transforms pullbacks over $D' \hookrightarrow D$ in $\Set^{\rightarrow}$ to pullbacks over $(D, D')$ in $\Mono$. Thus, suppose we have two objects $f, g$ of $\Set^{\rightarrow}$ with degree morphisms to $D' \hookrightarrow D$:
```
ScriptRaccoon: > pairs $(X, X')$ where $X' \subseteq X$

Let's try to find a consistent notation. It was (X,S) in the introduction.

ScriptRaccoon: What are "degree morphisms"?

databases/catdat/data/categories/Mono.yaml
ScriptRaccoon: I think the proof is easier with the description of set-subsets pairs.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+
+      We easily see that $\im(X \times_{D'} Y \to Z \times_D W) \subseteq \im(f) \times_{D'} \im(g)$ (this is the inclusion that we automatically get from the canonical comparison morphism). Conversely, suppose we have $z\in \im(f)$ and $w\in \im(g)$ with the same degree (which must be in $D'$). Then there exist $x\in X$ such that $f(x) = z$ and $y\in Y$ such that $g(y) = w$. Since $\deg_X(x) = \deg_Z(z) = \deg_W(w) = \deg_Y(y) \in D'$, we have $(x, y) \in X \times_{D'} Y$, and the image of $(x, y)$ in $Z \times_D W$ is exactly $(z, w)$, finishing the proof of the claim.
+
+      Therefore, from Lemma 5 <a href="/content/subcategories">here</a>, we conclude that each slice of $\Mono$ is cartesian closed, i.e. that $\Mono$ is locally cartesian closed.
+
+  - property: regular subobject classifier
+    proof: 'The object $\{ \top, \bot \} \xhookrightarrow{\id} $\{ \top, \bot \}$ is a regular subobject classifier, with $\top : (1, 1, \id) \to (\{ \top, \bot \}, \{ \top, \bot \}, \id)$ consisting of the maps with image $\top$. This follows from the description of regular monomorphisms below.'
+
+  - property: locally finitely presentable
+    proof: >-
```
ScriptRaccoon, line 68: Let's add a short alternative proof like: "Alternatively, we already know that the category is cocomplete, and the objects of the form ... are finitely presentable and generate the category under filtered colimits".

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+      $$\begin{CD}
+      (X, X') @> f >> (Y, Y') \\
+      @V g VV @VV i_1 VV \\
+      (Z, Z') @>> i_2 > ((Y \setminus f(X)) \sqcup Z, i_1(Y') \cup i_2(Z')).
+      \end{CD}$$
+      Here $i_1$ sends $f(x) \mapsto g(x) \in Z$ for $x \in X$, and otherwise it sends $y \mapsto y$ for $y \in Y \setminus f(X)$; and $i_2$ is the inclusion of $Z$. We certainly have $i_2$ is injective; it remains to show $i_2^{-1}(i_1(Y') \cup i_2(Z')) = Z'$. However, we have
+      $$i_2^{-1}(i_1(Y') \cup i_2(Z')) = i_2^{-1}(i_1(Y')) \cup i_2^{-1}(i_2(Z'))$$
+      and obviously $i_2^{-1}(i_2(Z')) = Z'$. Therefore, it suffices to show $i_2^{-1}(i_1(Y')) \subseteq Z'$. Thus, suppose we have $z \in Z$ such that $i_2(z) \in i_1(Y')$. That means there exists $y' \in Y'$ such that $z = i_1(y')$. But in order for $i_1(y')$ to land in the $Z$ portion of $(Y \setminus f(X)) \sqcup Z$, we must have $i_1(y') \in f(X)$. That means there exists $x \in X$ such that $f(x) = y'$; and by the assumption on $f$, in fact $x \in X'$. Therefore, $i_2(g(x)) = i_1(f(x)) = y' = i_2(z)$, so $z = g(x) \in Z'$.
```
ScriptRaccoon: Typo:

```suggestion
      and obviously $i_2^{-1}(i_2(Z')) = Z'$. Therefore, it suffices to show $i_2^{-1}(i_1(Y')) \subseteq Z'$. Thus, suppose we have $z \in Z$ such that $i_2(z) \in i_1(Y')$. That means there exists $y' \in Y'$ such that $z = i_1(y')$. But in order for $i_1(y')$ to land in the $Z$ portion of $(Y \setminus f(X)) \sqcup Z$, we must have $y' \in f(X)$. That means there exists $x \in X$ such that $f(x) = y'$; and by the assumption on $f$, in fact $x \in X'$. Therefore, $i_2(g(x)) = i_1(f(x)) = y' = i_2(z)$, so $z = g(x) \in Z'$.
```

ScriptRaccoon: I think the last sentence

`Therefore, $i_2(g(x)) = i_1(f(x)) = y' = i_2(z)$, so $z = g(x) \in Z'$.`

is not quite correct. Instead:

`Therefore, $i_2(g(x)) = i_1(f(x)) = i_1(y') = z = i_2(z)$, so $z = g(x) \in Z'$.`

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+
+  - property: co-Malcev
+    proof: >-
+      For this proof we will work in the equivalent category of pairs $(X, X')$ where $X' \subseteq X$. Thus, suppose we have a coreflexive corelation $p : (X \sqcup X, X' \sqcup X') \twoheadrightarrow (E, E')$ with coreflexivity morphism $r : (E, E') \to (X, X')$. From the assumption that $p$ is an epimorphism, we have that $p : X \sqcup X \to E$ is surjective. Since $\Set$ is co-Malcev, it follows that $E \simeq X \sqcup_Y X$ for some subset $Y \subseteq X$. It remains to show that $E' = i_1(X') \cup i_2(X') \subseteq X \sqcup_Y X$. Certainly, since we have a morphism $(X \sqcup_Y X, i_1(X') \cup i_2(X')) \to (E, E')$ induced by $p$, we must have $i_1(X') \cup i_2(X') \subseteq E'$. On the other hand, any element of $E'$ is equal to either $i_1(x)$ or $i_2(x)$ for $x \in X$. In the first case, we must have $r(i_1(x)) = x \in X'$, so $i_1(x) \in i_1(X')$; and similarly for the second case.
```
ScriptRaccoon: Let's add a link to the page for **Set**.

ScriptRaccoon: Let's use \cong for isomorphism, not \simeq.

ScriptRaccoon: > `On the other hand, any element of $E'$ is equal to either $i_1(x)$ or $i_2(x)$ for $x \in X$.`

Maybe we can add something like `any element of $E'$ is an element of $E$ and hence ...`

ScriptRaccoon: > `In the first case, we must have $r(i_1(x)) = x \in X'$``

When I read a chain of relations like this $a = b \in T$, I always read it like: We know $a=b$ and $b \in T$, hence $a \in T$. But here, it's different: We know $a \in T$ and $a=b$, *hence* $b \in T$. Thus, I think it is better to write it as $b = a \in T$.

So here: $x = r(i_1(x)) \in X'$.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,184 @@
...
+
+  - property: effective cocongruences
+    proof: 'See the proof that $\Mono$ is co-Malcev: It shows that in fact any coreflexive corelation is equivalent to an effective cocongruence $X \sqcup X \to X \sqcup_Y X$.'
```
ScriptRaccoon: Do you think it's worth adding a property "every reflexive relation is effective" to the database? (Not in this PR of course.)

I think we have seen this a couple of times already (Haus, ...).

ScriptRaccoon (changes requested):

databases/catdat/data/categories/Mono.yaml
dschepler: It's just the way I tend to think of objects of a slice category: as an object of the base category equipped with a "degree" morphism to the target object defining the slice. That way - similar to e.g. the multiplication and inverse functions on a group - the map shifts somewhat into the background, but remains available under a predictable name for when it's needed. (Or in the terminology of Rocq or Lean, adapted and extended from Haskell, the map to the target object would become a "typeclass instance". And that, in turn, is roughly the Haskell version of what in many other programming languages would be called "operator overloading" or "function overloading". In case any of those analogies might help.)

So, there's nothing special about a "degree morphism", other than that it's designated as serving a particular purpose or role. Just like there's nothing intrinsically distinguishing the multiplication operation on a semigroup from a generic associative binary operation on a set.

I'd still be open to any suggestions on how to better explain this, preferably without resorting to just choosing random names for each map to the base and then having to look up what name means what while reading the argument.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,158 @@
...
+
+      Therefore, any finite coproduct of $(1, 0)$ and $(1, 1)$ is finitely presentable; this means any object of the form $([n], [m])$ with $m, n \in \IN$ and $m \le n$ is finitely presentable. All such objects form a small set. Moreover, any object of $\Mono$ is a filtered colimit of the subobjects $(F, F \cap X')$ where $F$ is a finite subset of $X$, and each $(F, F \cap X')$ is isomorphic to some $([n], [m])$.
```
ScriptRaccoon: Suggestion removes an empty line after the diagram and makes a formula displaystyle. Also, adds the formula for an object in Mono in the end.

```suggestion
      It is equivalent to the category of models of the finite limit sketch with a single limit cone
      $$\begin{CD}
      X @> \id >> X \\
      @V \id VV @VV f VV \\
      X @>> f > Y.
      \end{CD}$$
      Alternately, we already know $\Mono$ is cocomplete. We claim that both $(1, 0)$ and $(1, 1)$ are finitely presentable objects of $\Mono$. For the first, $(1, 0)$ represents the forgetful functor $(X, X') \mapsto X$; and this functor has a right adjoint $X \mapsto (X, X)$ so it preserves filtered colimits (in fact all colimits). For the second, $(1, 1)$ represents the functor $(X, X') \mapsto X'$. Now by the previous description of colimits, for a filtered diagram $((X_i, X_i'))_{i\in\I}$, first form the filtered colimit of $X_i$ in <a href="/category/Set">$\Set$</a> with cocone $c_i : X_i \to \colim_{i\in\I} X_i$. Then the filtered colimit in $\Mono$ is $(\colim_{i\in\I} X_i, \bigcup_{i\in\I} c_i(X_i'))$. This makes it easy to see that the canonical function
      $$\colim_{i\in\I} \Hom((1,1), (X_i,X_i')) \to \Hom((1,1), \colim_{i\in\I} (X_i,X_i'))$$
      is surjective. To see it is injective, use the fact that the unique morphism $(1, 0) \to (1, 1)$ is an epimorphism (see below), and we have already seen that the corresponding function for $(1,0)$ is injective.

      Therefore, any finite coproduct of $(1, 0)$ and $(1, 1)$ is finitely presentable; this means any object of the form $([n], [m])$ with $m, n \in \IN$ and $m \le n$ is finitely presentable. All such objects form a small set. Moreover, any object $(X,X')$ of $\Mono$ is a filtered colimit of the subobjects $(F, F \cap X')$ where $F$ is a finite subset of $X$, and each $(F, F \cap X')$ is isomorphic to some $([n], [m])$.
```

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,158 @@
...
+      $$\begin{CD}
+      (X, X') @> f >> (Y, Y') \\
+      @V g VV @VV i_1 VV \\
+      (Z, Z') @>> i_2 > ((Y \setminus f(X)) \sqcup Z, i_1(Y') \cup i_2(Z')).
+      \end{CD}$$
+      Here $i_1$ sends $f(x) \mapsto g(x) \in Z$ for $x \in X$, and otherwise it sends $y \mapsto y$ for $y \in Y \setminus f(X)$; and $i_2$ is the inclusion of $Z$. We certainly have $i_2$ is injective; it remains to show $i_2^{-1}(i_1(Y') \cup i_2(Z')) = Z'$. However, we have
+      $$i_2^{-1}(i_1(Y') \cup i_2(Z')) = i_2^{-1}(i_1(Y')) \cup i_2^{-1}(i_2(Z'))$$
+      and obviously $i_2^{-1}(i_2(Z')) = Z'$. Therefore, it suffices to show $i_2^{-1}(i_1(Y')) \subseteq Z'$. Thus, suppose we have $z \in Z$ such that $i_2(z) \in i_1(Y')$. That means there exists $y' \in Y'$ such that $z = i_1(y')$. But in order for $i_1(y')$ to land in the $Z$ portion of $(Y \setminus f(X)) \sqcup Z$, we must have $y' \in f(X)$. That means there exists $x \in X$ such that $f(x) = y'$; and by the assumption on $f$, in fact $x \in X'$. Therefore, $i_2(g(x)) = i_1(f(x)) = i_1(y') = z = i_2(z)$, so $z = g(x) \in Z'$.
```
ScriptRaccoon: Displaystyle for long equation:

```suggestion
      and obviously $i_2^{-1}(i_2(Z')) = Z'$. Therefore, it suffices to show $i_2^{-1}(i_1(Y')) \subseteq Z'$. Thus, suppose we have $z \in Z$ such that $i_2(z) \in i_1(Y')$. That means there exists $y' \in Y'$ such that $z = i_1(y')$. But in order for $i_1(y')$ to land in the $Z$ portion of $(Y \setminus f(X)) \sqcup Z$, we must have $y' \in f(X)$. That means there exists $x \in X$ such that $f(x) = y'$; and by the assumption on $f$, in fact $x \in X'$. Therefore,
      $$i_2(g(x)) = i_1(f(x)) = i_1(y') = z = i_2(z),$$
      so $z = g(x) \in Z'$.
```

databases/catdat/data/categories/Mono.yaml
ScriptRaccoon: I see. There can be an advantage of using such a terminology when one needs these morphisms over and over again, but here, for an isolated proof, there is no harm to add simple letters to denote them. You can also just always use $p_X : X \to S$ for the structure morphism of an object in a slice category. (Any other latter than $p$ also works.)

The main problem with the formulation "with degree morphisms" is that it is no standard term, and I am afraid that most readers will be confused by it. It would be different if you introduce them properly (like "morphisms ... which we denote by ... because ... ").

databases/catdat/data/categories/Mono.yaml
ScriptRaccoon: **Lemma.** Let $U$ be a functor which is right adjoint to a faithful functor $F$. Then $U$ preserves cogenerators.

*Proof.* If $Q$ is a cogenerator, $Hom(-,U(Q)) \cong Hom(F(-),Q) \cong Hom(-,Q) \circ F$ is a composition of faithful functors, hence faithful. QED

**Corollary.** The category **Mono** has a cogenerator.

Proof: Apply the Lemma to $U : Set \to Mono$, $U(X) = (X,X)$, which is right adjoint to $F : Mono \to Set$, $F(X,X') = X$, and $F$ is faithful.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,158 @@
+id: Mono
+name: category of sets with a distinguished subset
+notation: $\Mono$
```
ScriptRaccoon, line 3: I really like the decision to change the presentation of the category. But now, the notation is a bit off, don't you think?

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,158 @@
...
+morphisms: >-
+  a morphism $(X, X') \to (Y, Y')$ is a function $f : X \to Y$ such that $f(X') \subseteq Y'$
+description: >-
+  This is equivalent to the full subcategory of objects $(X, Y, f)$ of <a href="/category/Set_arrow">$\Set^{\rightarrow}$</a> where $f : X \to Y$ is an injective function.
+nlab_link: https://ncatlab.org/nlab/show/M-category#def
+
+tags:
+  - set theory
+
+related_categories:
```
ScriptRaccoon: With the new presentation of this category, we might even add something like **Pos** as a related category. A partial order is a set together with a binary relation + properties. An object of this cartegory is a set together with a unary relation. But I am not 100% sure if this makes sense.

databases/catdat/data/categories/Mono.yaml
ScriptRaccoon, line 1: Do you also want to add the functors to the database which appear in this discussion? For example, the functor Mono ---> Set which maps (X,X') to X.

databases/catdat/data/categories/Mono.yaml
ScriptRaccoon: In case I merge https://github.com/ScriptRaccoon/CatDat/pull/269 first:

```suggestion
related:
```

dschepler:
At this point, if you decide you prefer a different notation, I'm happy with whatever you settle on.

ScriptRaccoon:
What about **Sub**?

dschepler:
> What about **Sub**?

I'd be worried about it being too easy to confuse with the subobject functor on a well-powered category with pullbacks.

I was thinking **UnaryPred** could be an option (category of sets with a unary predicate), which would be similar to **BinRel** for category of sets with a binary relation.

databases/catdat/data/categories/Mono.yaml
ScriptRaccoon: I assume this is a No. :D
I will do this myself in another PR.

ScriptRaccoon:
> I'd be worried about it being too easy to confuse with the subobject functor on a well-powered category with pullbacks.

True.

> I was thinking **UnaryPred** could be an option (category of sets with a unary predicate), which would be similar to **BinRel** for category of sets with a binary relation.

Wouldn't **UnaryRel** then be more similar?

But the name is a bit technical, I like **Mono** more actually. Let's just keep it for now; and we can change it if (for some reason) we add the "true" **Mono** later.

ScriptRaccoon:
On the other hand: What is the standard notation for the category of pairs (X,A) of a topological space X and a subspace A? This appears in homology theories. [Mathlib](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Category/TopPair.html?utm_source=chatgpt.com) uses **TopPair**. So this should be **SetPair**? But I don't like this, since it can be easily confused with **Set** &times; **Set**. [These notes](https://studyres.com/doc/1854462/homology-theory---section-de-math%C3%A9matiques?utm_source=chatgpt.com&__cf_chl_f_tk=rvTOsNkjyg0yUbQb4As3etB0sWXAwz7_.gRbkrPH0TQ-1783435254-1.0.1.1-wkabvhi05qSd.0o0vFY4zwwaqab7IEkyhbADN3OFsQg) use **Top**<sub>(2)</sub>. So we might use **Set**<sub>(2)</sub>? I mean it looks good, but it doesn't convey any specific meaning, and it could also stand for a category of pairs of sets $(X,Y)$ subject to some other condition. On the other hand, I also used random notation like **Set**<sub>f</sub> for [this category](https://catdat.app/category/Set_f).

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,160 @@
+id: Mono
+name: category of sets with a distinguished subset
+notation: $\Mono$
+objects: >-
+  pairs $(X, X')$ where $X$ is a set and $X' \subseteq X$ is a subset
+morphisms: >-
+  a morphism $(X, X') \to (Y, Y')$ is a function $f : X \to Y$ such that $f(X') \subseteq Y'$
+description: >-
+  This is equivalent to the full subcategory of objects $(X, Y, f)$ of <a href="/category/Set_arrow">$\Set^{\rightarrow}$</a> where $f : X \to Y$ is an injective function.
```
ScriptRaccoon:
```suggestion
  This is equivalent to the full subcategory of objects $(X, Y, f)$ of <a href="/category/Set_arrow">$\Set^{\rightarrow}$</a> where $f : X \to Y$ is an injective function, i.e. a monomorphism. This explains our notation.
```

(if we keep the notation)

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,160 @@
...
+      We have that $\Mono$ is a reflective subcategory of $\Set^{\rightarrow}$, with the reflector taking a function $f : X \to Y$ to the pair $(Y, \im(f))$. Therefore, the result follows from the fact that <a href="/category/Set_arrow">$\Set^{\rightarrow}$</a> is cocomplete.
+
+      More concretely, suppose we have a diagram $((X_i, X_i'))_{i \in \I}$, with colimit cone $c_i : X_i \to \colim_{i\in\I} X_i$ in $\Set$. Then the colimit in $\Mono$ is $(\colim_{i\in\I} X_i, \bigcup_{i\in\I} c_i(X_i'))$.
+    check_redundancy: false
+
+  - property: disjoint coproducts
+    proof: This follows from the fact that coproducts in $\Mono$ can be computed component-wise, along with the fact that <a href="/category/Set">$\Set$</a> has disjoint coproducts.
+
+  - property: locally cartesian closed
+    proof: >-
```
ScriptRaccoon, line 55: Alternative (I let you decide which proof you prefer): **Set**<sup>&rarr;</sup> is locally cartesian closed because it is a Grothendieck topos. The general construction from [insert source] shows that the exponential of ... is given by [insert description]. Since **Mono** is a full subcategory of **Set**<sup>&rarr;</sup> which is closed under all (finite) limits, it remains to check that it is closed under exponentials. [insert proof for this]

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,160 @@
...
+  regular monomorphisms:
+    description: >-
+      morphisms $f : (X, X') \to (Y, Y')$ such that $f$ is an injective function from $X$ to $Y$, and $f^{-1}(Y') = X'$ (in particular, if $f : X \to Y$ is an inclusion map, then this is equivalent to $X' = X \cap Y'$)
+    proof: >-
+      To show any regular monomorphism satisfies this condition, suppose we have an equalizer diagram
+      $$\begin{CD}
+      (X, X') @> f >> (Y, Y') @> g > h > (Z, Z').
+      \end{CD}$$
+      We need to show $f^{-1}(Y') = X'$. From the definitions, we automatically get $X' \subseteq f^{-1}(Y')$. For the other inclusion, suppose we have $x \in X$ such that $f(x) \in Y'$. Then certainly $g(f(x)) = h(f(x)) \in Z'$, so $x'$ is in the equalizer of $g |_{Y'}, h |_{Y'} : Y' \rightrightarrows Z'$, which by definition is exactly $X'$.
```
ScriptRaccoon: What is the element $x'$? 

I guess this is a typo.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,158 @@
...
+    check_redundancy: false
+
+  - property: cocomplete
+    proof: >-
+      We have that $\Mono$ is a reflective subcategory of $\Set^{\rightarrow}$, with the reflector taking a function $f : X \to Y$ to the pair $(Y, \im(f))$. Therefore, the result follows from the fact that <a href="/category/Set_arrow">$\Set^{\rightarrow}$</a> is cocomplete.
+
+      More concretely, suppose we have a diagram $((X_i, X_i'))_{i \in \I}$, with colimit cone $c_i : X_i \to \colim_{i\in\I} X_i$ in <a href="/category/Set">$\Set$</a>. Then the colimit in $\Mono$ is $(\colim_{i\in\I} X_i, \bigcup_{i\in\I} c_i(X_i'))$. (It is easy to see that in the subset component, we can also restrict $i$ to range over a <a href="https://ncatlab.org/nlab/show/weak+multilimit" target="_blank">weakly initial</a> set of objects of $I$.)
```
ScriptRaccoon: I think it must be a weakly terminal set.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,158 @@
...
+
+  - property: cogenerator
+    proof: >-
+      Consider the forgetful functor $U : \Mono \to \Set$, $(X, X') \mapsto X$. This has right adjoint $R : \Set \to \Mono$, $X \mapsto (X, X)$. Therefore, by the dual of Lemma 9 <a href="/content/subcategories">here</a>, $R$ preserves cogenerators; and in particular, since $\Set$ has a cogenerator, so does $\Mono$.
+
+  - property: complete
+    proof: >-
+      The component-wise limit gives a pair of a set and a subset, and this pair forms the limit in $\Mono$.
+
+      More concretely, suppose we have a diagram $((X_i, X_i'))_{i \in \I}$, with limit cone $p_i : \lim_{i\in\I} X_i \to X_i$ in <a href="/category/Set">$\Set$</a>. Then the limit in $\Mono$ is $(\lim_{i\in\I} X_i, \bigcap_{i\in\I} p_i^{-1}(X_i'))$. (It is easy to see that in the subset component, we can also restrict $i$ to range over a <a href="https://ncatlab.org/nlab/show/weak+multilimit" target="_blank">weakly terminal</a> set of objects of $I$.)
```
ScriptRaccoon: I think it must be a weakly initial set.

databases/catdat/data/categories/Mono.yaml
```diff
@@ -0,0 +1,158 @@
...
+    proof: >-
+      For the non-trivial direction, use the fact that the forgetful functor $\Mono \to \Set$, $(X, X') \mapsto X$, has a right adjoint given by $X \mapsto (X, X)$. Therefore, the forgetful functor preserves epimorphisms; so if $f : (X, X') \to (Y, Y')$ is an epimorphism in $\Mono$, then $f : X \to Y$ is an epimorphism in <a href="/category/Set">$\Set$</a> and therefore surjective.
+  regular monomorphisms:
+    description: >-
+      morphisms $f : (X, X') \to (Y, Y')$ such that $f$ is an injective function from $X$ to $Y$, and $f^{-1}(Y') = X'$ (in particular, if $f : X \to Y$ is an inclusion map, then this is equivalent to $X' = X \cap Y'$)
+    proof: >-
+      To show any regular monomorphism satisfies this condition, suppose we have a parallel pair $g, h : (Y, Y') \to (Z, Z')$. To form the equalizer in $\Mono$, first we form the equalizer $f : X \hookrightarrow Y$ in $\Set$ of the underlying functions $g$ and $h$. Then by the construction of limits in the proof of completeness above, the equalizer in $\Mono$ is $(X, f^{-1}(Y'))$. Thus, we see that $f^{-1}(Y') = X'$.
```
ScriptRaccoon: `g, h : (Y, Y') \rightrightarrows (Z, Z')` (like for regular epis below)

ScriptRaccoon (approved):
Looks good now. Thank you for the PR! 🚀

PR #274, opened 2026-07-09
Check that environment variables are up-to-date

ScriptRaccoon:
When a new environment variable is added or an existing one is removed, this change is currently only reflected in `.env.example`, since `.env` is not tracked in the repository. As a result, contributors may end up with a broken development server without any notification that their local `.env` file needs to be updated.

This PR adds a check to detect this situation.

Whenever the development server is started, a script verifies that the environment variables defined in `.env` match those in `.env.example`. If they differ, an error is thrown with a message asking the developer to update their `.env` file.

The script intentionally does not run with the test or build commands. These commands run in CI, where no `.env` file exists; environment variables are provided via GitHub secrets instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
 	"version": "0.0.1",
 	"type": "module",
 	"scripts": {
-		"dev": "vite dev",
+		"dev": "pnpm env:check && vite dev",
 		"dev:remote": "vite dev --mode prod",
 		"dev:host": "vite dev --host",
 		"build": "vite build",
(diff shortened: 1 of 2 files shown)


ScriptRaccoon:
WDYT @dschepler ? Is this a good solution?

dschepler:
> WDYT @dschepler ? Is this a good solution?

It seems fine to me.

ScriptRaccoon:
Thank you for the feedback. Then I will merge.

ScriptRaccoon:
This PR didn't age well ...
In https://github.com/ScriptRaccoon/CatDat/pull/278 the check is removed again. Reason: All environment variables are now public anyway, so `.env` can now be in git.

Issue #22, opened 2026-03-27
Deduce properties based on other categories


ScriptRaccoon:
If a category has coproducts, then each of its slices also has coproducts. A product of elementary topoi is also an elementary topos. There are many results of this form. Currently, when using such a result, it is used "ad hoc". It would be great if we could

- add results like these to the database (as separate entities)
- use these results to deduce properties (and non-properties) automatically

varkor:
> What is more doable: mark those properties that descend to subcategories. Also, mark those that descend to full subcategories.

More generally, might it be possible to support implications of the following form?

> Given a functor F : X -> Y, if X has properties P, Y has properties Q, and F has properties R, then X (respectively Y) has properties S.

As well as being more flexible in the code, it would also permit properties like "creation of limits implies completeness of the domain", etc.

ScriptRaccoon:
Yes, this is the best way to generalize this. Interestingly, then we need to run deductions in "multiple rounds". Currently they are done with categories, then with functors. But by using the implications you mentioned, we need to run them again with categories. But then we need to run them again with functors, etc. Probably there is no bound, one just has to stop when no new property can be assigned.

ScriptRaccoon:
A small remark that puts this issue into perspective:

For a given category, we manually assign some properties and then deduce many others. For that reason, the category detail page distinguishes between manual and deduced assignments.

However, the property detail page has no such distinction. When a category is assigned to a property (a perspective not used so far in CatDat), no additional categories are deduced.

So this issue is essentially an attempt to treat categories and properties more equally when it comes to assignments.

PR #301, opened 2026-07-18
Add proof dependencies (WIP)

ScriptRaccoon:
## Why we need proof dependencies

Many proofs of properties depend on other properties. While the deduction system takes care of many proofs of this type, some proofs cannot be done automatically via implications. In these cases, this information was stored only in the proof text and was not validated at all.

For example, if a proof referred to the "fact" that **Grp** is abelian, no contradiction would be detected, even though the database of course contains the information that **Grp** is not abelian. It might also happen that a proof uses the fact that a certain category has a property, even though the database does not contain this information at all, i.e. the property has not been decided yet.

The mentioned issues can easily be prevented with a good review process. However, the worst case is that we discover that some proof is wrong and cannot be fixed. In that case, we remove the assignment (or even negate it!), but we cannot be sure whether any other proof depends on the removed assignment or how to locate these proofs. This is a non-trivial problem, since the proofs might depend on properties that were deduced from the deleted assignment. (This is actually a big problem in the mathematical literature, which cannot be fixed in that context.)

This PR fixes this by checking that all proof dependencies are valid. As a necessary step, all proof dependencies now need to be documented.

The feature has been suggested by Daniel in [this comment](https://github.com/ScriptRaccoon/CatDat/pull/231#discussion_r3375851213).

## How

Concretely, for every property assignment that depends on another property assignment, this dependency must be added manually alongside the assignment. Here is a typical example from `Alg(R).yaml`:

```yaml
  - property: cofiltered-limit-stable epimorphisms
    proof: We already know that <a href="/category/CAlg(R)">$\CAlg(R)$</a> does not have this property. Now apply the contrapositive of the dual of Lemma 2 <a href="/content/subcategories">here</a> to the forgetful functor $\CAlg(R) \to \Alg(R)$. It preserves epimorphisms by <a href="https://math.stackexchange.com/questions/5133488" target="_blank">MSE/5133488</a>.
    dependencies:
      - id: CAlg(R)
        type: category
        property: cofiltered-limit-stable epimorphisms
        satisfied: false
```

The seed script writes these dependencies to a new table in the database called `required_property_assignments`. The checks cannot be performed at this point since some requirements refer to deduced properties. (For example, the proof that **Cat** is infinitary extensive depends on the fact that **Set** is infinitary extensive, and the latter is *deduced* from the fact that it is a Grothendieck topos.) Thus, we first let the deduction script run and only perform the checks in the test script (`pnpm db:test`). There, a new function `test_proof_dependencies` checks whether all claimed properties are actually valid.

## Multiple dependencies

Some proofs refer to several other properties. This is also supported. For example, this is the proof that the walking fork is locally cartesian closed:

```yaml
  - property: locally cartesian closed
    proof: We need to check that every slice category is cartesian closed. The slice category over $0$ is the <a href="/category/1">trivial category</a>. The slice category over $1$ is the <a href="/category/walking_morphism">walking morphism</a>. Finally, the slice category over $2$ ist the <a href="/category/walking_commutative_square">walking commutative square</a>. All of these are cartesian closed, see their pages for details.
    dependencies:
      - id: '1'
        type: category
        property: cartesian closed
        satisfied: true
      - id: walking_morphism
        type: category
        property: cartesian closed
        satisfied: true
      - id: walking_commutative_square
        type: category
        property: cartesian closed
        satisfied: true
```

## References are dependencies

References to facts are treated as dependencies as well, since otherwise they do not make any sense. For example, here is a passage from `CAlg(R)` (with a proof excerpt):

```yaml
  - property: regular quotient object classifier
    proof: 'The strategy is similar to the one for <a href="/category/CRing">$\CRing$</a>: Assume that [...]'
    dependencies:
      - id: CRing
        type: category
        property: regular quotient object classifier
        satisfied: false
        with_proof: true
```

Here is another example from `CAlg(R)`:

```yaml
  - property: Malcev
    proof: This follows in the same way as for <a href="/category/Grp">$\Grp$</a>, see also Example 2.2.5 in <a href="https://ncatlab.org/nlab/show/Malcev,+protomodular,+homological+and+semi-abelian+categories" target="_blank">Malcev, protomodular, homological and semi-abelian categories</a>.
    dependencies:
      - id: Grp
        type: category
        property: Malcev
        satisfied: true
        with_proof: true
```

## Referenced proofs

Sometimes, not just properties are referenced, but proofs are referenced. This means that the assignments need to be non-deduced. To mark these references, the property `with_proof: true` can be used.

Here is an example from `Man.yaml` (with a proof excerpt).

```yaml
  - property: ℵ₁-cofiltered limits
    proof: We continue the counterexample from <a href="/category/Set_c">$\Set_c$</a>. There, we [...]
    dependencies:
      - id: Set_c
        type: category
        property: ℵ₁-cofiltered limits
        satisfied: false
        with_proof: true
```

This is important because we might remove the direct proof for **Set**<sub>c</sub> once we find an implication that handles this case. If we did so, the proof for **Man** would no longer make sense.

In fact, this PR has identified such a case: the proof that **Top**<sub>\*</sub> has no regular quotient object classifier refers to the proof that **Set**<sub>\*</sub> has none. However, the latter proof was removed some time ago (in #77), so the proof for **Top**<sub>\*</sub> no longer makes sense. EDIT: I have fixed this now in a separate commit in the main branch: https://github.com/ScriptRaccoon/CatDat/commit/4041ddb4244252d4babadac859e5349e8c2c7516

## Technical stuff

We need to add the `type` information (which is currently either `category`, `functor`, or `morphism`) to each dependency since a property is no longer uniquely determined by its name (cf. #263). One *could* derive the type from the `id`, which still identifies structures, but for now I have decided not to do that.

## TODOs

1. I have skipped all proofs for (co)generators and (co)generating sets because of the on-going PR #280 and it is not clear which PR gets merged first. Check this later.
2. Mark dependencies with `with_proof: true` if appropriate. EDIT: No, use proof IDs instead. But this seems clumsly ...
3. Fix missing referenced proofs

## Next

Another issue is to prevent circularity. This can be checked in another PR.

This work can also be seen as a preparation for #22.


ScriptRaccoon:
I think this PR has become way too complex and needs to be split down into separate, smaller PRs. The first one is #319 (and maybe that feature is already sufficient).

Issue #267, opened 2026-07-05
Add more types of morphisms


ScriptRaccoon:
Continuing https://github.com/ScriptRaccoon/CatDat/pull/265, which introduced morphisms as a new structure type and added a couple of types of epimorphisms already, more types of epimorphisms should be added:

| Property | ✓ | PR |
| --- | --- | --- |
| [effective epimorphism](https://ncatlab.org/nlab/show/effective+epimorphism) | ✅ | #291
| [strict epimorphism](https://ncatlab.org/nlab/show/strict+epimorphism) | ✅ | #291
| [extremal epimorphism](https://ncatlab.org/nlab/show/extremal+epimorphism) | ✅| #320 
| [strong epimorphism](https://ncatlab.org/nlab/show/strong+epimorphism) | ✅ | #296
| [normal epimorphism](https://ncatlab.org/nlab/show/normal+epimorphism) | ✅ | #292
| [descent morphism](https://ncatlab.org/nlab/show/descent+morphism) | | 
| [effective descent morphism](https://ncatlab.org/nlab/show/effective+descent+morphism) | | 

The corresponding types of monomorphisms should also be added.

Also:

| Property | ✓ | PR |
| --- | --- | --- |
| [constant morphism](https://ncatlab.org/nlab/show/constant+morphism) | ✅ |  #332
| [coconstant morphism](https://math.stackexchange.com/questions/158270/intuition-for-coconstant-morphisms) | ✅ | #332 
[strict morphism](https://ncatlab.org/nlab/show/strict+morphism) ||

Also, add the corresponding results that compare these properties.

Bonus: If the page with missing data lists consistent yet unwitnessed combinations, add examples of morphisms. For example, a regular epimorphism which is not effective, etc.

ykawase5048:
How about adding "being left (right) orthogonal to all mono(epi)morphisms" to the property list? If the category has binary powers, this implies epimorphicity (i.e., strong epimorphism), and if the category has pullbacks, it coincides with being an extremal epimorphism. However, in general, it is a slightly different notion from either of them.

PR #320, opened 2026-08-04
Add extremal monomorphisms and extremal epimorphisms

ScriptRaccoon:
This PR fills another gap from #267 by adding extremal monomorphisms and extremal epimorphisms.

Some basic results are added as well (e.g. that every strong mono is extremal, and that the converse holds when pushouts exist).

## The "forked commutative square"

Since every consistent morphism combination (or its dual) should have at least one witness, this PR also adds an example of a non-strong extremal monomorphism. Apparently, such examples are not so easy to find "in nature", which is why this PR also adds a new category to the database: the "forked commutative square", which is generated by the graph

<br />

and the evident relations. Here, $m$ is an extremal monomorphism, but not a strong one. All properties have been decided for this category (and for $m$). The most interesting part is the (long) proof that it has cosifted limits.

*EDIT.* Meanwhile, another example has been found by Daniel at [MSE](https://math.stackexchange.com/questions/5146052/). I have added this as well.

## Special morphism rules

Moreover, three special morphism rules have been added:

- in a gaunt category, isomorphisms are the same as identities (by definition)
- in a regular-subobject-trivial category, regular monos are the same as isomorphisms (by definition)
- in a regular-quotient-trivial category, regular epis are the same as isomorphisms (by definition)

This made it easier to classify the special morphisms in the forked commutative square, but it also means that several isomorphism assignments in examples of gaunt categories could be removed.

## Graphical overview

This PR also adds a content page that includes a graphical overview of the relationships between the various types of epimorphisms and monomorphisms.

<br />

<br />

Maybe this is better suitable for the nLab, and I tried to add this to the page on epimorphisms, but the nLab keeps throwing errors at me. So let's keep it here for now. Of course, it should be adjusted as soon we add more types of epimorphisms or monomorphisms (such as descent morphisms).


ScriptRaccoon:
@dschepler I think I am finished here. Can you please review this PR? (Probably you are still busy with the PR on total categories, which means that, of course, this here can wait.)

ScriptRaccoon:
Sorry for the back and forth here. I had asked for a review, but I'll go ahead and merge this now to keep things moving. There are also a few other open PRs that need attention, and some of them are starting to depend on each other.

PR #322, opened 2026-08-07
Add symmetric monoidal categories

### LINKED ISSUE #318: Add support for symmetric monoidal categories
Currently, CatDat supports three types of categorical structures: categories, functors, morphisms. This issue is about adding [symmetric monoidal categories](https://ncatlab.org/nlab/show/symmetric+monoidal+category). It is a follow-up of #4. See #290 for a similar issue.

Every symmetric monoidal category should have an underlying category which is recorded in the database. 

Some examples (but these can be added over time, not necessarily in the first PR):
- $(Set,\times,1)$
- $(Top,\times,1)$
- $(Set,\sqcup,0)$
- $(Cat,\times,1)$
- $(Set_*,\wedge,\\{0,1\\})$
- $(Ab,\otimes,Z)$
- $(Ab_{fg},\otimes,Z)$
- $(Mod_R,\otimes_R,R)$ (where $R$ is a commutative ring)
- $(grMod_R, \otimes_R,R)$ with the symmetry $\sigma(a \otimes b) = b \otimes a$
- $(grMod_R, \otimes_R,R)$ with the symmetry $\sigma(a \otimes b) = (-1)^{|a| \cdot |b|} \cdot b \otimes a$
- $(Sh(X,Ab),\otimes,\underline{\mathbb{Z}})$
- $(N,+)$
- $([0,\infty],+)$
- $(Ban,\otimes_p,C)$

Some properties:

- cartesian
- cocartesian
- closed
- strict
- distributive (i.e. the underlying category has coproducts and $\otimes$ preserves coproducts in each variable)
- cocomplete (i.e. the underlying category is cocomplete and $\otimes$ preserves colimits in each variable)

(!) Properties that only refer to the underlying category are *not* added to the database. There is no need.

The mentioned properties should be decided for the mentioned symmetric monoidal categories.

If $(C,\otimes,1)$ is a symmetric monoidal category, then $(C^{op},\otimes,1)$ is a symmetric monoidal category; dual properties refer to this kind of dualization. The other kind of dualization, using $\otimes^{op}$, only becomes relevant for monoidal categories, which are not the subject here.

Technical detail: for URLs, the type should be with underscores `symmetric_monoidal_category`, but for text, it should be with spaces `symmetric monoidal category`. Thus, one needs to differentiate between these two use cases (which was not necessary before). Maybe this can be a separate PR before adding symmetric monoidal categories.

As a follow-up, symmetric monoidal functors can be added as well.

ScriptRaccoon:
This PR implements symmetric monoidal categories and thus resolves #318.

## Features

As for other categorical structures, users can switch to symmetric monoidal categories, view examples and properties of symmetric monoidal categories, look for implications, and perform property combination searches and comparisons.

## Added symmetric monoidal categories

- $(Set,\times,1)$
- $(Set,\sqcup,0)$
- $(Cat,\times,1)$
- $(Top,\times,1)$
- $(\mathbb{N},\leq,+,0)$
- $(Ab,\otimes,\mathbb{Z})$
- $(Ab_{fg},\otimes,\mathbb{Z})$
- $(R{-}Mod,\otimes,R)$ (when $R$ is a non-zero commutative ring)
- $(R{-}Mod,\otimes,R)$ (when $R$ is absolutely flat)
- $(R{-}Mod,\otimes,R)$ (when $R$ is not absolutely flat)
- $(FinVect_K,\otimes,K)$
- $(1,\times,0)$

There are the following parent-child relationships (cf. #323):

- $(Ab,\otimes,\mathbb{Z})$ is a child of $(R{-}Mod,\otimes,R)$ where $R$ is not absolutely flat
- $(R{-}Mod,\otimes,R)$ (where $R$ is absolutely flat) and $(R{-}Mod,\otimes,R)$ (where $R$ is not absolutely flat) are both children of $(R{-}Mod,\otimes,R)$ (where $R$ is a non-zero commutative ring) 

Of course, more symmetric monoidal categories should be added in the future. This is just the first batch.

## Added properties of symmetric monoidal categories

- cartesian / cocartesian
- closed / coclosed
- cocomplete* / complete
- finitely cocomplete / finitely complete
- distributive / codistributive
- infinitary distributive / infinitary codistributive
- well-pointed
- strict
- trivial
- self-dual

*This means that the underlying category is cocomplete and $\otimes$ preserves colimits in each variable. There is no overlap with the notion of a cocomplete category as long as we carefully distinguish between a symmetric monoidal category and its underlying category. Hence, there is no need for a more complicated term such as "monoidally cocomplete". (Technically, storing two properties with the same name but different types has been made possible by #263.)

## Added implications

There are too many to list here. Just to mention three examples:

- finitely cocomplete ===> distributive
- closed + underlying category is cocomplete ===> cocomplete
- distributive + underlying category has biproducts ===> codistributive

## Proofs

The mentioned properties have been determined for the symmetric monoidal categories listed above, either by direct proofs, implications, or inheritance from parent categories.


ScriptRaccoon:
@dschepler  Do you agree with my decision (and the reasoning in the PR and [this file](https://github.com/ScriptRaccoon/CatDat/pull/322/changes#diff-0983a33c24eefe3d07715933c9a1827351d09051ec09dc9951a4e6b2ddf882eb)) that "cocomplete" suffices as a property of symmetric monoidal categories and that we do not need a term like "monoidally cocomplete"?

PR #337, opened 2026-08-16
Justify mathematical texts

ScriptRaccoon:
This PR implements justified text in property proofs (in popups), implication pages, descriptions, and content pages by using the CSS declaration `text-align: justify`. Justified text is generally not recommended on the web for accessibility reasons ([article on this matter](https://theadminbar.com/accessibility-weekly/text-alignment-left-center-justified/)), but it is more consistent with the usual layout of mathematical writing, such as in journals, on the arXiv, and on the nLab, and the negative effect of excessive spacing between words is mitigated by enabling hyphenation throughout the body. Headings are excluded from justification and hyphenation.

### Before

1-before

### After

1-after

### Before

2-before

### After

2-after

### Before

3-before

### After

3-after

### Before

4-before

### After

4-after

## Issues

Despite the hyphenation, some sentences still have excessively large spaces between words.

spaces<br />

This seems to happen when the browser does not know how to hyphenate a technical term. In this case, one can use a soft hyphen: replacing `subterminal` with `sub&shy;terminal` fixes the issue. However, this is not really scalable.

We also still have the non-trivial issue of formulas being broken across lines (#135).

break<br />

More generally, the text needs to remain readable at every viewport width. This is a significant difference from the layout of a classical paper.

### DIFF
--- a/content/constant_morphisms.md
+++ b/content/constant_morphisms.md
@@ -30,7 +30,7 @@ If $X$ is a subterminal object, then any morphism $X \to Y$ is constant. If $Y$ 
 _Proof._ This is immediate from the definitions. <span class="qed">$\square$</span>
 
 ::: Lemma 5
-If $f : X \to Y$ is a monomorphism that is constant, then $X$ is subterminal.
+If $f : X \to Y$ is a monomorphism that is constant, then $X$ is sub&shy;terminal.
 :::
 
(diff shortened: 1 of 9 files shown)


ScriptRaccoon:
Actually I have just decided against this after looking at this paragraph (from `foundations.md`). The white space is annoying. And it cannot be controlled at all (in a scalable way).

Bildschirmfoto 2026-08-16 um 14 44 46

Arguably, most texts look better with `text-align: justify` and `hyphens: auto`, but some do not, and it has a [negative effect on a11y](https://theadminbar.com/accessibility-weekly/text-alignment-left-center-justified/). Also, justification does not solve a concrete *problem*.

PR #362, opened 2026-09-08
Add the effective topos

### LINKED ISSUE #353: Add effective topos
An idea I've been toying around with is to see what it would take to add the effective topos to the database: https://ncatlab.org/nlab/show/effective+topos

For example, I think this is an example of an elementary topos with natural numbers object which does not have countable copowers (and in fact the natural numbers object is not a countable copower of the terminal object).

I'm not sure how well I'd be able to assign other properties, though. I think I've just barely gotten to the point where I can grasp the basic definitions of what an object is and what a morphism is, and I can start to reason about the category.

(Anyway, my immediate plans would be to finish off the quasitopos project by adding a property for "has effective regular congruences" - the grouping is "has effective (regular congruences)" rather than "has (effective and regular) congruences". Then add the Grothendieck quasitopos property, and then maybe add the category SepPsh(X) for X a topological space. And then possibly a few of the other examples we've discussed, such as "category of sets equipped with a binary relation", "category of subsequential spaces", "category of pseudotopological spaces" (that one I think would be an example of a complete and cocomplete quasitopos which is not a Grothendieck quasitopos because it doesn't have a small extremal generating set), "category of simple graphs".)

dschepler:
Addresses #353

Currently unresolved properties:
is [accessible](http://localhost:5173/category-property/accessible)
is [coaccessible](http://localhost:5173/category-property/coaccessible)
has a [cogenerating set](http://localhost:5173/category-property/cogenerating_set)
has a [cogenerator](http://localhost:5173/category-property/cogenerator)
is [concretizable](http://localhost:5173/category-property/concretizable)
has an [extremal cogenerating set](http://localhost:5173/category-property/extremal_cogenerating_set)
has an [extremal cogenerator](http://localhost:5173/category-property/extremal_cogenerator)
has an [extremal generating set](http://localhost:5173/category-property/extremal_generating_set)
has an [extremal generator](http://localhost:5173/category-property/extremal_generator)
has a [generating set](http://localhost:5173/category-property/generating_set)
has a [generator](http://localhost:5173/category-property/generator)
is [ℵ₁-accessible](http://localhost:5173/category-property/aleph1-accessible)
has [ℵ₁-cofiltered limits](http://localhost:5173/category-property/aleph1-cofiltered_limits)
has [ℵ₁-filtered colimits](http://localhost:5173/category-property/aleph1-filtered_colimits)


ScriptRaccoon:
Thanks! I will have a look.

Is the PR ready for review already?

It would be cool to prove that **Eff** has no generating set (which has been claimed [here](https://math.stackexchange.com/questions/649829/generators-in-a-topos)). Currently, CatDat does not have an example of an elementary topos that does not have a generating set.

dschepler:
I think it's probably mostly ready for review - though there is one spot where we still need to find a reference (or else expand to a full proof of the elementary topos properties).

At the link, my understanding was the claim is that the effective topos doesn't have a small dense subcategory. I'd certainly believe that since I think I can prove the only coproducts which exist are the ones where all but finitely many terms are empty. It might be possible to extend that to a proof that the only colimits which exist are the ones which are more or less equivalent to a finite colimit, but I'm not sure about a formal formulation or proof of that.

That would still, of course, be sufficient to prove that it is not accessible, and it looks like we currently don't have an example of an elementary topos which is not accessible.

ScriptRaccoon:
> though there is one spot where we still need to find a reference 

Yes and I think it's a reference for the most important property. Please try to locate it.

dschepler:
I've found a reference https://www.sciencedirect.com/science/article/pii/S0022404905001593 which claims in passing "it is relatively easy to see" the collection of countable projectives is a generating set for Eff - though I don't see the "relatively easy" proof myself.

ScriptRaccoon:
> I've found a reference https://www.sciencedirect.com/science/article/pii/S0022404905001593 which claims in passing "it is relatively easy to see" the collection of countable projectives is a generating set for Eff - though I don't see the "relatively easy" proof myself.

In such cases, we can already add a proof to CatDat with the reference, along with a `TODO` comment to explain this more directly since the proof is incomplete. CatDat also aims to record what is known in the literature even if we do not understand everything.

I have not begun to read the changes in this PR, but I will do that tomorrow.

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+  - property: natural numbers object
+    proof: 'We claim that $N \coloneqq (\IN, \{ (n, n, n) \mid n \in \IN \})$ is a natural numbers object. To see this, suppose we have an object $(X, E_X)$ with morphisms $x : 1 \to X$ and $f : X \to X$. Then we can find $n_0 \in \IN, x_0 \in X$ such that $n_0 \Vdash 0 \xmapsto{x} x_0$; also, let $c$ and $t$ be witnesses to the codomain property and totality of $f$, respectively. We can then define an "iteration of $c\circ t$" program encoded by $i\in\IN$, such that $i \cdot n$ represents the result of the $n$-fold iteration of $t$ followed by $c$, starting with $c \cdot n_0$. We can also find a sequence $x_n \in X$ extending $x_0$ such that $t \cdot (i \cdot n) \Vdash x_n \xmapsto{f} x_{n+1}$. We can then show that $\{ (\langle n, m \rangle, n, y) \mid (m \Vdash x_n = y) \}$ is an effective functional relation which induces a morphism $N \to X$, and this morphism has the required properties. Likewise, the uniqueness of this morphism is straightforward to establish.'
+
+  - property: generating set
```
ScriptRaccoon, line 58: ℹ️ This needs to be changed to `generating collection` after #367 is merged.  The proof does not change much.

database/data/category-implications/topos.yaml
ScriptRaccoon, line 1: Let's add a reference that the power set functor is monadic (in each case). It is Theorem 2.2.7 in Johnstone.

You can copy paste it from `power_set_contravariant.yaml`

database/data/categories/Eff.yaml
ScriptRaccoon, line 1: ℹ️ Once you rebase on main and run `pnpm db:update`, this will add the effective topos to a JSON file (`structure.history.json`). Please commit the change.

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
+id: Eff
+name: effective topos
+notation: $\Eff$
+objects: A set $X$ along with a subset $E_X \subseteq \IN \times X \times X$ giving an effective partial equivalence relation (see below for precise definition)
+morphisms: A morphism $(X, E_X) \to (Y, E_Y)$ is a subset $f \subseteq \IN \times X \times Y$ giving an effective functional relation, modulo a certain equivalence relation (see below for precise definitions)
+description: >-
```
ScriptRaccoon, line 6: This description is very detailed and maybe should be moved to the nLab. The other entries also refer for more detailed descriptions to the nLab (or other sources).

I did something similar with the category http://localhost:5173/category/cocompletion-discrete-pair-join, but this was a different situation since this category does not even deserve an nLab article.

If you agree to move it, of course, the description here does not need to repeat it. But please apply the changes suggested in the other comments first.

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+  To give a precise definition, first fix a certain Turing-equivalent theory of computation, with an encoding of programs as elements of $\IN$ (note that not every element of $\IN$ needs to be a valid program encoding). We also fix a pairing function $\langle {-}, {-} \rangle : \IN \xrightarrow{\sim} \IN \times \IN$ along with the inverse pair $p_1, p_2 : \IN \rightrightarrows \IN$. We will define notation $p \cdot n \downarrow$ to mean that $p$ is a valid program encoding, and running the corresponding program with input $n$ terminates without error, returning a value $p \cdot n$.
```
ScriptRaccoon, line 9: What do you mean by "along with the inverse pair"? Isn't N -> N x N already the same direction? Or did you want to write N x N -> N?

EDIT: yeah from the rest it is clear that you mean N x N -> N. Please fix the direction.

ScriptRaccoon, line 9: > encoding of programs

Which programs?

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+  Given a subset $E_X \subseteq \IN \times X \times X$, we will use notation $n \Vdash x_1 = x_2$ (or informally, $n$ realizes equality of $x_1$ and $x_2$) to mean $(n, x_1, x_2) \in E_X$. We will say that $E_X$ is an <i>effective partial equivalence relation</i> if $(p_1)_*(E_X)$ is a partial equivalence relation on $X$ with partial recursive functions relating the corresponding realizers; namely:
```
ScriptRaccoon, line 11: How is $(p_1)_*(E_X)$ even a relation on $X$? What is your $p_1$? Obviously it cannot be the first projection.

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+  * (Symmetry:) There exists $s \in \IN$ such that whenever $n \Vdash x_1 = x_2$, then $s \cdot n \downarrow$ and $s \cdot n \Vdash x_2 = x_1$.
+
+  * (Transitivity:) There exists $t \in \IN$ such that whenever $n \Vdash x_1 = x_2$ and $m \Vdash x_2 = x_3$, then $t \cdot \langle n, m \rangle \downarrow$ and $t \cdot \langle n, m \rangle \Vdash x_1 = x_3$.
+
```
ScriptRaccoon, line 16: Why is reflexivity missing?

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+  Likewise, suppose $f \subseteq \IN \times X \times Y$ where we have fixed effective partial equivalence relations $E_X$ and $E_Y$ on $X$ and $Y$, respectively. We then use notation $n \Vdash x \xmapsto{f} y$ to mean $(n, x, y) \in f$. We will say $f$ is an <i>effective functional relation</i> if $(p_1)_*(f)$ is a functional relation $X / E_X \to Y / E_Y$, with partial recursive functions relating the corresponding realizers; namely:
```
ScriptRaccoon, line 17: As before, the notation `(p_1)_*(f)` is unclear to me. Maybe you mean `(p_{2,3})_*(f)`.

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+  * (Domain condition:) There exists $d \in \IN$ such that whenever $n \Vdash x \xmapsto{f} y$, then $d \cdot n \downarrow$ and $d \cdot n \Vdash x = x$.
```
ScriptRaccoon, line 19: What is the intuition behind this condition? Why do we need it?

database/data/categories/Eff.yaml
ScriptRaccoon, line 11: How can a natural number model an equality x = y? What is the picture in the computing analogy? Is the natural number an encoding of a proof?

And what is $E_X$ then, intuitively?

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+  * Conversely, there exists $\ell_2$ such that whenever $n \Vdash x \xmapsto{g} y$, then $\ell_2 \cdot n \downarrow$ and $\ell_2 \cdot n \Vdash x \xmapsto{f} y$.
+
+  The effective topos is then defined to be the category whose objects are pairs $(X, E_X)$ where $X$ is a set and $E_X$ is an effective partial equivalence relation on $X$; and the morphisms $\Hom((X, E_X), (Y, E_Y))$ are the quotient of the set of effective functional relations by the equivalence relation defined above. The identity morphism of $(X, E_X)$ is the equivalence class of $E_X$; in other words, it is $[\id]$ where $n \Vdash x \xmapsto{\id} y$ if and only if $n \Vdash x = y$. Composition is defined such that for $f$ an effective functional relation between $(Y, E_Y)$ and $(Z, E_Z)$ and $g$ an effective functional relation between $(X, E_X)$ and $(Y, E_Y)$, $[f] \circ [g]$ is the equivalence class of $f \circ g \coloneqq \{ (\langle n, m\rangle, x, z) \mid \exists y\in Y, (n \Vdash x \xmapsto{g} y) \land (m \Vdash y \xmapsto{f} z) \}$.
+
+nlab_link: https://ncatlab.org/nlab/show/effective+topos
+
+tags:
+  - computation theory
```
ScriptRaccoon, line 40: This tag is a bit specific. I suggest to take "logic".

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+  * (Image uniqueness condition:) There exists $u \in \IN$ such that whenever $n \Vdash x \xmapsto{f} y_1$ and $m \Vdash x \xmapsto{f} y_2$, then $u \cdot \langle n, m \rangle \downarrow$ and $u \cdot \langle n, m \rangle \Vdash y_1 = y_2$.
+
+  * (Totality condition:) There exists $t \in \IN$ such that whenever $n \Vdash x = x$, then $t \cdot n \downarrow$, and there exists $y \in Y$ such that $t \cdot n \Vdash x \xmapsto{f} y$.
+
+  Similarly, for two such functional relations $f$ and $g$, we will say $f$ and $g$ are equivalent if $(p_1)_*(f) = (p_1)_*(g)$, with partial recursive functions relating the corresponding realizers; namely:
+
+  * There exists $\ell_1$ such that whenever $n \Vdash x \xmapsto{f} y$, then $\ell_1 \cdot n \downarrow$ and $\ell_1 \cdot n \Vdash x \xmapsto{g} y$.
+
+  * Conversely, there exists $\ell_2$ such that whenever $n \Vdash x \xmapsto{g} y$, then $\ell_2 \cdot n \downarrow$ and $\ell_2 \cdot n \Vdash x \xmapsto{f} y$.
```
ScriptRaccoon, line 33: Why is this relation reflexive and transitive?

For reflexivity we probably need an "identity program"?

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+  The effective topos is then defined to be the category whose objects are pairs $(X, E_X)$ where $X$ is a set and $E_X$ is an effective partial equivalence relation on $X$; and the morphisms $\Hom((X, E_X), (Y, E_Y))$ are the quotient of the set of effective functional relations by the equivalence relation defined above. The identity morphism of $(X, E_X)$ is the equivalence class of $E_X$; in other words, it is $[\id]$ where $n \Vdash x \xmapsto{\id} y$ if and only if $n \Vdash x = y$. Composition is defined such that for $f$ an effective functional relation between $(Y, E_Y)$ and $(Z, E_Z)$ and $g$ an effective functional relation between $(X, E_X)$ and $(Y, E_Y)$, $[f] \circ [g]$ is the equivalence class of $f \circ g \coloneqq \{ (\langle n, m\rangle, x, z) \mid \exists y\in Y, (n \Vdash x \xmapsto{g} y) \land (m \Vdash y \xmapsto{f} z) \}$.
```
ScriptRaccoon, line 35: Why is composition well-defined? Why are the two category axioms satisfied? Please add a reference. (I think the proofs are too long to be written down here. Maybe on a content page it would work, or the nLab article.)

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+related:
+  - Set
+
+satisfied_properties:
+  - property: locally small
+    proof: This is trivial.
+
+  - property: elementary topos
+    proof: This is a special case of the fact that from any <a href="https://ncatlab.org/nlab/show/tripos">tripos</a> we can construct a topos. See for example <a href="https://www.cl.cam.ac.uk/~amp12/papers/thet/thet.pdf">A. M. Pitts, <i>The Theory of Triposes</i></a>, with the definition of effective tripos in section 1.5 on p. 16, and the construction of a topos corresponding to a tripos as Thm. 2.3.
```
ScriptRaccoon, line 50: It is a bit unfortunate that readers, in particular those who are not familiar with the whole theory (like myself), after reading the long and complicated definition of the effective topos, which even left out some details, now stumble upon another notion, that of a **tripos** and need to understand it from the references. The relation is also not immediately clear to me. Which tripos do we use here? And why do we get the effective topos in that example? More importantly, why are we even using the long-winded definition above then?

I will stop reading for now, since I first need to understand more basic things before being able to understand any of the following proofs.

database/data/categories/Eff.yaml
```diff
@@ -0,0 +1,171 @@
...
+    proof: 'There is a fully faithful (and therefore conservative) functor $\Set \to \Eff$ which sends $X$ to $(X, \IN \times \Delta_X)$, and which sends $f : X \to Y$ to the equivalence class of $\IN \times \Gamma_f$ where $\Gamma_f$ is the graph of $f$.'
+
+  - property: locally finite
+    proof: For example, $\Hom(1, N)$ is countably infinite where $N$ is the natural numbers object.
+
+special_objects:
+  initial object:
+    description: $(\varnothing, \varnothing)$
+  terminal object:
+    description: $(1, \{ (0, 0, n) \mid n \in \IN \})$
```
ScriptRaccoon, line 95:
```suggestion
    description: $(1, \{ (n, 0, 0) \mid n \in \IN \})$
```

database/data/categories/Eff.yaml
dschepler: Programs of the chosen Turing-equivalent theory of computation (e.g. Turing machines, or untyped lambda calculus expressions, etc.).

database/data/categories/Eff.yaml
dschepler: That's intentional. A value witnessing $x = x$ is meant to represent a certificate that $x$ is a "valid" member of the type. It could sort of be viewed as a parallel to forcing, where a name of a set in the forcing model includes conditions under which certain things are added as elements of the set.

database/data/categories/Eff.yaml
dschepler: The relation shouldn't provide a value of the function for an $x\in X$ which isn't even considered a valid member of the type. So, more closely, this is saying there is a way to take a "certificate" that the function maps $x$ to $y$, and to return a certificate that $x$ is a valid value.

database/data/categories/Eff.yaml
dschepler: In order to form a topos, you need to be able to take quotients of effective equivalence relations. Without the quotients, what you get are called $\omega$-sets -- or "assemblies of the first Kleene combinatory algebra" or "assemblies" for short if clear from context -- and that category looks like it's only a quasitopos (I saw a result in one of the articles I browsed that it's in fact equivalent to the $\lnot\lnot$-separated objects of **Eff**). So, we also allow for "certificates" of some sort that two values are to be considered equivalent.

(And then, allowing for partial equivalence relations which for some $x\in X$ have no witnesses at all that $x$ is valid is a matter of convenience for some of the constructions.)

Maybe it could help if I replace $n \Vdash x = y$ with $n \Vdash x \equiv y$?  And perhaps use some special notation for $n \Vdash x \equiv x$ such as $n \Vdash x \downarrow$ to emphasize that we're considering it as a sort of "validity predicate"?

database/data/categories/Eff.yaml
dschepler: Right - for example, if you chose untyped lambda calculus as your Turing-equivalent theory of computation, then you'd just use an encoding of $\lambda x . x$.

database/data/categories/Eff.yaml
dschepler: I could certainly move it to content/ . I found the descriptions at nLab to be very abstract and spread across multiple pages, which is part of why it took me so long to decode the definitions into this more concrete form for this specific case - so I'm not sure if a reference to nLab would be useful for giving the definitions.

database/data/categories/Eff.yaml
dschepler: You have a point - similarly to the way I found I needed to digest the abstract definitions into a more concrete definition for this case, it would probably be useful to write up this special case of the proofs as well.

database/data/categories/Eff.yaml
ScriptRaccoon: I see. I see nLab as the major online reference for category theory. CatDat should *not* be competition. I want that CatDat is the place that records structured data of categories and other categorical structures. When a comprehensive and self-contained definition of the effective topos is missing, this is material for nLab. More people will find it (via google etc.) and profit from it. Either the existing article can be edited, or a separate article can be created (like Todd did for ETCS years ago).

dschepler: I'm not sure about that. I don't see many nLab pages that have detailed proofs of everything, more often pages with overviews of the important properties and results, with maybe a few select proofs to be highlighted.

Though I could certainly register for an nLab account, start off by editing the "quasitopos" page with the correction of "generating set" to "strong generating set" (in nLab terminology) that we discovered a while ago, and then see if I can start a discussion of whether and/or where it could be appropriate to add such a thing.

In the meantime, I think it might help to start with the category of $\omega$-sets, which uses a lot of the same ingredients but the definitions are simpler. (In that category, an object is a set $X$ along with a "realizes" relation which is a subset of $\mathbb{N} \times X$ such that every $x \in X$ is realized by some $n\in \mathbb{N}$, and a morphism is a function $f : X \to Y$ such that there exists a partial recursive function $\varphi : \mathbb{N} \to \mathbb{N}$ such that whenever $n$ realizes $x$, then $\varphi(n) \downarrow$ and $\varphi(n)$ realizes $f(x)$. Then the effective topos could be viewed as freely adding quotients of congruences; even though the category of $\omega$-sets does have quotients of congruences already, they aren't effective quotients in general, and re-adding quotients turns out to make those quotients effective and result in a topos.)

ScriptRaccoon: > I'm not sure about that. I don't see many nLab pages that have detailed proofs of everything

That is true. But some pages have very long and detailed proofs, too. But notice that I was only suggesting to put the detailed *definition* to the nLab. The only required proof is that the result is a category (thus, answering my other questions).

> Though I could certainly register for an nLab account

No account is required. You can click on "Edit" and go ahead. Just enter your name in the textbox and submit. 👍🏻

> In the meantime, I think it might help to start with the category of ω-sets

This sounds like a really good idea.

database/data/categories/Eff.yaml
ScriptRaccoon: Ok. In any writeup of the whole thing (either as content page or, what I would prefer, nLab page), things like this deserve a whole paragraph. Here, where a category should be defined briefly, it needs to be omitted, I am afraid.

database/data/categories/Eff.yaml
ScriptRaccoon: Ok. Yeah adding $\omega$-sets probably will clarify lots of things.

Yes, I prefer $n \models x \equiv y$ over $n \models x = y$. And when you introduce that notation, it will be definitely helpful to also explain what you did just here.

I don't like to have another notation for $n \models x \equiv x$, but it will be helpful to explain in 1-2 sentences what it means intuitively.

database/data/category-implications/topos.yaml
```diff
@@ -107,6 +107,22 @@
...
+    - elementary topos
+    - generator
+  conclusions:
+    - cogenerator
+  proof: 'For an elementary topos $\T$, the power object functor $P : \T^{\op} \to \T$ is right adjoint to $P^{\op} : \T \to \T^{\op}$ and moreover is faithful. The first statement is part of the proof that $P$ is monadic (<a href="https://ncatlab.org/nlab/show/Sketches+of+an+Elephant" target="_blank">Johnstone</a>, 2.2.7) and the second is a corollary. Therefore, from Lemma 9 <a href="/content/subcategories"> here, if $G$ is a generator of $\T$, then $PG$ is a cogenerator.'
+
+- id: topos_generating_set_implies_cogenerating_set
+  assumptions:
+    - elementary topos
+    - generating set
```
ScriptRaccoon, line 121: *generating collection

I assume this is also why the tests fail.

database/data/categories/Eff.yaml
dschepler: OK, I've created a new branch dschepler/CatDat/omega-set and I'll start working on populating that for now.

PR #366, opened 2026-09-10
Redesign home page

ScriptRaccoon:
The design of the home page has been significantly improved. Users can now get a better sense of how CatDat can be used, while directly seeing recorded examples of categories, functors, and other entries in the database.

## Key metrics

Cards displaying the key metrics of the database have been added. Some example entries are also highlighted and linked directly.

key-metrics<br />

The numbers are animated, starting from 0. The animation is only shown once per session.
 
## Recently added structures

The five most recently added structures are displayed, giving the database a more dynamic and evolving feel. The list can also be expanded to show 10 structures.

recent<br />

To make this possible, the seed script now writes newly seeded structures to a JSON file when they are added. The information cannot be stored in the database since this is always recreated from scratch.

## Features

The feature cards have been replaced with text sections. More examples are provided for the search and comparison features, allowing users to explore them directly from the home page.

features<br />

## YouTube video

The YouTube video iframe has been removed. The video is outdated anyway, and it is better to use the home page to explain the features.

## Navigation

The "Home" link from the navigation menu has been removed. This is because the logo already links to the home page (a change from #348). More importantly, the user will probably not go to the homepage anyway after seeing other detail pages.

## Before / After

before&nbsp;&nbsp;&nbsp;after<br />


shared/structure.history.json
ScriptRaccoon, line 1: The dates are not precise:

1. Many categories were added already in January of 2025 when I started to code CatDat, but back then, they were recorded in a file `001_categories.sql`, and only later separate files were created. The creation dates of these individual files* has been used to compile this JSON file.

2. Some category IDs simply had been renamed (e.g. `Set_f` to `Set_ff`).

But as for the purpose of this file, this is OK. I only want to show the 5 (resp. 10) most recent additions on the home page. This is why older historical data is irrelevant.

*or rather, their property files, see https://github.com/ScriptRaccoon/CatDat/commit/3e6b46c3cd62fefdc8d29c957736fefb3f4d5756

Issue #83, opened 2026-04-15
How to handle smallness?


ScriptRaccoon:
This is a continuation of https://github.com/ScriptRaccoon/CatDat/issues/25 and the discussions in #59 about the notion of a small set and a small category. I want to have this issue focussed on just one property, namely being small. We can generalize later.

## The problem

The question is basically: do we want to have the category property of being "small" in the database? Currently, we have it. But being a set or not depends on set theoretic constructions. The proofs that a category is small or not may depend on choices that are completely irrelevant from a category theoretic perspective. Here is a comment by @ykawase5048  from the linked PR:

> However, to be honest, I feel a little reluctance about assigning such properties, which are not invariant under isomorphisms. The reason is that, in order to determine whether a category is small, one must fix a set-theoretic construction of that category, but I'm not convinced that such information can always be extracted from the webpage of each category in CatDat. For example, the terminal category is small if its collections of objects and morphisms are constructed as, for example, { ∅ }. But, it might be constructed in a more pathological way, for instance as the functor category [Set,1]. In that case, it would no longer be small.

The underlying issue is that being small (as it is currently defined) is not invariant under isomorphisms.

And that is essentially because a collection which is isomorphic to a set does not have to be a set.

Here is another comment by @ykawase5048 :

> As I mentioned in https://github.com/ScriptRaccoon/CatDat/issues/25, I still believe that the term "essentially small" is confusing. Even though up-to-iso smallness and up-to-equiv smallness coincide for discrete categories, I still think that using "essentially small" for up-to-iso notions goes against the intuition of many category theorists. In particular, when a collection of objects in a category is referred to, category theorists often tend to identify the collection with the full subcategory spanned by those objects. Then, "up-to-iso small" and "up-to-equiv small" are no longer the same concept.

> Compromise. It might be better to use "small family" instead in this case. Here, the smallness refers to the index set, so it naturally means up-to-iso smallness. (But, strictly speaking, a family is not the same as a collection, since it allows duplication of elements.)

> [...]

> Let me take the opportunity to restate my thoughts about smallness here: I still believe that up-to-iso smallness of categories should be included in the database, and that it should share the same name as up-to-iso smallness of sets. However, the term "essentially small" conflicts with the existing notion of up-to-equiv smallness of categories.

[...]

Issue #25, opened 2026-03-30
Local smallness vs. locally essential smallness


ykawase5048:
I recommend deleting the property "locally small" in the current sense, because it is not invariant even under isomorphisms of categories: the naive smallness of a set is not preserved by bijections.
I think such a non-invariance shouldn't be included in the database.

Moreover, as well as I know, many category theorists use "small" to mean "isomorphic to a (naively) small one". In this usage, a category is locally small precisely when each hom-class is isomorphic to an element in the fixed (Grothendieck) universe, which is precisely the current definition of “locally essentially small".
So, I also suggest renaming "locally essentially small" to simply "locally small".

ykawase5048:
> Additionally, the term "locally essentially small" may be misleading, since it can sound like each hom-category of a 2-category (or bicategory) is essentially small, i.e., equivalent to some small category.

@ScriptRaccoon Does this make sense to you? (Though, this is just my personal impression.)

ykawase5048:
Yes, and for the same reason, I think we should also remove "small", as it is also not invariant under isomorphism, and instead add the up-to-iso smallness.
But, the name "essentially small"’ is already used to mean up-to-equivalence smallness ...

ScriptRaccoon:
> Yes, and for the same reason, I think we should also remove "small", as it is also not invariant under isomorphism, and instead add the up-to-iso smallness. But, the name "essentially small"’ is already used to mean up-to-equivalence smallness ...

This would create confusion. Also, small categories are quite important.

I think we need to accept the fact that categories are used in two ways, both "structurally" and "combinatorially", and the latter usage *does not* care if properties are invariant under isomorphisms of categories.

> If 2-categories are added to the database eventually (cf. https://github.com/ScriptRaccoon/CatDat/issues/4), then there will be ambiguity.

No, since properties of 2-categories will be completely separate things as properties of 1-categories (separate tables in the db etc.), even when they have the same name. This is already happening now: a "small functor" will never be confused with a "small category".

>  Furthermore, in 2-category theory it is not uncommon to implicitly view categories as locally discrete 2-categories.

Yes, but this will not happen in CatDat, since every relationship will be made *explicit*.

I will close this issue now for the reasons I mentioned. In a nutshell, CatDat should speak about the properties that appear in the literature and not create any new conventions.

EDIT: follow-up issue is #83

jonas-frey:
I agree that "small" should be stable under iso. I think that reflects mathematical practice. 

[...]

PR #367, opened 2026-09-10
Consistent usage of essentially small collections

ScriptRaccoon:
CatDat works with set-theoretic foundations based on sets and collections. Sets are also called small collections. But in many situations, collections appear that are only bijective to a set, either implicitly or explicitly. This calls for a term for such collections and a consistent usage of this term.

I went with "essentially small collection" in this PR, since it is consistent with the definition of a locally essentially small category (which was added months ago) and also because a discrete category is essentially small if and only if its collection of objects is essentially small in this sense. The page on foundations (`/content/foundations`) has been improved accordingly.

The usage of "small set" has been removed, since it is not compatible with the foundations we are using, or rather, it leaks an implementation detail (collections are also sets in a certain universe, just renamed). Instead, we should speak of sets, aka small collections.

While going through all usages of the term "small" in the repository, some usages turned out to be redundant, while others turned out to be wrong. 

An example appears in the definition of an accessible category, where it was stated that there "is a (small) set of objects G such that ...". First of all, it is confusing (regardless of what we are working with here) to use parentheses in a definition, since this raises the unanswered question of whether the word in parentheses is required or not. As mentioned, "small set" is redundant and can simply be abbreviated to "set".

But a more important problem is that this definition of an accessible category is not entirely correct. A set has sets as elements. This means that the objects in G are sets. But there are accessible categories in nature whose objects (which must only be collections) are not sets. For example, it can happen that they are only isomorphic to sets. But it is not correct here anyway to go "inside" the structure of the objects. The definition of an accessible category should merely require that G is an *essentially small collection* of objects. (Equivalently, it can be "parametrized" by a set.) This has been fixed:

> Let $\kappa$ be a regular cardinal. A category is <i>$\kappa$-accessible</i> if it has $\kappa$-filtered colimits and there is an essentially small collection $G$ of $\kappa$-presentable objects such that every object is a $\kappa$-filtered colimit of objects in $G$. A category is <i>accessible</i> if it is $\kappa$-accessible for some regular cardinal $\kappa$.

For similar reasons, the terminology "generating set" was not quite correct. This notion is now called a "generating collection", and it is essentially small by convention (since otherwise every category would have one). It is a bit unfortunate that we cannot simply use the term "generating set", which is much more common in the literature, but it is not correct since the objects of a category are not necessarily sets, so an essentially small collection of objects is not necessarily a set. The same changes apply to the related properties, which are now called: "extremal generating collection", "cogenerating collection", and "extremal cogenerating collection".

In issues on set-theoretic questions (#25, #83), my suggestion of an "essentially small collection" and of a "locally essentially small category" (which was added long ago) has been criticized because an essentially small collection of objects in a category does not necessarily constitute a full subcategory that is essentially small (namely, when the ambient category is not locally essentially small). To this, I would just respond that the identification between a collection of objects and a full subcategory should be made explicit, and this avoids any confusion. If we clearly speak, for example in the definition of an accessible category, about a collection of objects, I am convinced that no confusion can arise.

Another criticism was that "essentially small" already means "equivalent to a small category". But this is only true for categories. The new term applies to a completely different type of mathematical object, so there is no overlap as long as we distinguish the types properly, which we should do anyway. Compare this with the notions "exact category" and "exact functor", which cannot be confused as long as we do not identify a category with its identity functor.

Other set-theoretic formulations have been improved as well. For example, since diagrams and hence their limits are small by convention unless otherwise stated, it is redundant to write "small products", or even worse, "(small) products".

This PR was motivated by working on #365. For such categories, precise set-theoretic foundations are required.


ScriptRaccoon:
FYI @varkor  @dschepler. To summarize the most important changes:
- a collection isomorphic to a set is now called "essentially small"
- the term "small set" is banned
- "generating set" is now called "generating collection", and it is essentially small by convention (*not a set*)

⚠️ The objects of a category are not necessarily sets. Hence, an ess. small collection of objects is not necessarily a set.

PR #373, opened 2026-09-15
Add category of ω-sets

dschepler:
As discussed in the PR for the effective topos (#353), this PR is working on putting in the subcategory of ω-sets (otherwise known as a special case of the category of assemblies for the partial combinatory algebra also used in constructing the effective topos). The subcategory has simpler definitions that are easier to work with, to work up to the full definition of the effective topos. It can also be useful in an alternate construction of the effective topos, as the ex/reg completion of this category.


dschepler:
Currently unresolved properties (many similar to the ones that were unresolved for the full effective topos):
is [accessible](http://localhost:5173/category-property/accessible)
is [Barr-coexact](http://localhost:5173/category-property/Barr-coexact)
is [coaccessible](http://localhost:5173/category-property/coaccessible)
has [effective cocongruences](http://localhost:5173/category-property/effective_cocongruences)
has an [extremal cogenerating collection](http://localhost:5173/category-property/extremal_cogenerating_collection)
has an [extremal cogenerator](http://localhost:5173/category-property/extremal_cogenerator)
has an [extremal generating collection](http://localhost:5173/category-property/extremal_generating_collection)
has an [extremal generator](http://localhost:5173/category-property/extremal_generator)
is [ℵ₁-accessible](http://localhost:5173/category-property/aleph1-accessible)
has [ℵ₁-cofiltered limits](http://localhost:5173/category-property/aleph1-cofiltered_limits)
has [ℵ₁-filtered colimits](http://localhost:5173/category-property/aleph1-filtered_colimits)

At this point, I think the best way to prove effective cocongruences would be to incorporate the proof from the abandoned quasitopos + SepPsh(X) PR, that regular + extensive + quotients of congruences -> effective cocongruences.

Otherwise, I might be able to adapt the proof from the effective topos PR to show that the $\omega$-sets with underlying sets contained in $\mathbb{N}$ are an extremal generating collection - though I haven't quite gotten the details on that down yet.

dschepler:
On reviewing it myself, I see a need to improve the wording where I refer to a partial recursive function $\mathbb{N} \dashrightarrow \mathbb{N}$ "being a map as in the definition" for a function between the underlying sets. Maybe we can come up with a good name for that partial recursive map - something like "a realizer tracking map for $f$" perhaps?

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+      (Note that in the case where $(Z, R_Z)$ is terminal, we can simplify the description of exponentials in $\omega{-}\Set$ as follows: We define $[(X, R_X), (Y, R_Y)]$ to be the set $\Hom_{\omega{-}\Set}((X, R_X), (Y, R_Y))$ with the realization relation that $p$ realizes $f$ if and only if the $p$th partial recursive function is unary $\IN \dashrightarrow \IN$ and it works in the definition to show that $f$ is a morphism of $\omega$-sets.)
+
```
dschepler, line 89: I'm not really happy with this proof as it stands -- there are a lot of moving parts to keep track of, and I get the feeling it will probably be confusing for many readers, especially the fourth paragraph on converting a map from the pullback to a generalized element of the relative exponential. I'm not sure at the moment what would be a good way to improve the presentation, though. So if you have any suggestions along those lines, they would certainly be welcome.

dschepler: Hmm... It might be possible to define a "regularly partial morphism classifier" $\tilde Y$ whose underlying set is $Y \sqcup \{ undef \}$ and whose realization relation is that everything realizes $undef$, while $n$ realizes $y$ in $\tilde Y$ if and only if $n$ is a code for a partial recursive function such that $n \cdot 0 \downarrow$ and $n \cdot 0$ is a realizer for $y$ in $Y$. That means a morphism $X \to \tilde Y$ should be equivalent to a regular subobject $X'$ of $X$ and a morphism $X' \to Y$. (The indirection in realizers is necessary in order to have something to do with realizers of elements in $X$ which aren't necessarily realizers of any elements of $X'$.) I should also be able to show directly that the category is cartesian closed, using the simplified definition, and then adapt one of the standard proofs that a topos is locally cartesian closed to this situation (constructing $[X, Y]_Z$ as an equalizer of two certain maps from $[X, \tilde Y] \times Z$).

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
+id: omega-Set
+name: category of ω-sets
+notation: $\omega{-}\Set$
+objects: pairs $(X, R_X)$ where $X$ is a set and $R_X \subseteq \IN \times X$ is a "realization" relation such that $(p_2)_*(R_X) = X$
+morphisms: 'a morphism $(X, R_X) \to (Y, R_Y)$ is a function $f : X \to Y$ such that there exists a <a href="https://ncatlab.org/nlab/show/partial+recursive+function">partial recursive function</a> $\varphi : \IN \dashrightarrow \IN$ such that whenever we have $n\in \IN, x\in X$ such that $n$ realizes $x$ (i.e. $(n, x) \in R_X$), then $\varphi(n)$ exists and $\varphi(n)$ realizes $f(x)$'
+description: >-
+  The category of $\omega$-sets is mostly useful as a <a href="https://ncatlab.org/nlab/show/dense+subcategory">dense subcategory</a> of the <a href="https://ncatlab.org/nlab/show/effective+topos">effective topos</a> consisting of objects which are simpler to work with. Namely, the effective topos can be constructed as the <a href="https://ncatlab.org/nlab/show/regular+and+exact+completions">ex/reg completion</a> of $\omega-{\Set}$.
```
ScriptRaccoon:
```suggestion
  The category of $\omega$-sets is mostly useful as a <a href="https://ncatlab.org/nlab/show/dense+subcategory">dense subcategory</a> of the <a href="https://ncatlab.org/nlab/show/effective+topos">effective topos</a> consisting of objects which are simpler to work with. Namely, the effective topos can be constructed as the <a href="https://ncatlab.org/nlab/show/regular+and+exact+completions">ex/reg completion</a> of $\omega{-}\Set$.
```

Alternative: create a macro, since this appears quite often here

ScriptRaccoon:
> On reviewing it myself, I see a need to improve the wording where I refer to a partial recursive function N ⇢ N "being a map as in the definition" for a function between the underlying sets. Maybe we can come up with a good name for that partial recursive map - something like "a realizer tracking map for f " perhaps?

I have no expertise on this category at all, but I would probably call it "a realization function of the morphism", or simply "a realizer of the morphism". Then you can describe morphisms as realizable functions.

*EDIT.* after reading some of the proofs (still on a superficial level), I think that any term will greatly simplify the exposition; it will be used all the time

database/data/categories/omega-Set.yaml
ScriptRaccoon: > I'm not really happy with this proof as it stands -- there are a lot of moving parts to keep track of, and I get the feeling it will probably be confusing for many readers, especially the fourth paragraph on converting a map from the pullback to a generalized element of the relative exponential. I'm not sure at the moment what would be a good way to improve the presentation, though. So if you have any suggestions along those lines, they would certainly be welcome.

I am not there yet to go through the mathematics of the proofs, but I have some suggestions regarding the layout, which often helps with understanding a proof (when done right).

Currently, the proof has a few paragraphs, but still, it is a long wall of text.

wall of text<br />

My suggestions are:

- use the vertical space generously (in particular after #316 !)
- use as much paragraphs as you like
- use displayed formulas more often
- for long or complex proofs, use html to add sections to the proof

Doing these adjustments, it looks much better and is easier to read, and probably also to understand.

structured proof<br />

You can find the adjusted proof here: [proof.txt](https://github.com/user-attachments/files/32308153/proof.txt)

This is just my first shot. It could even be improved much further (but for this, I would need to digest the proof).

Also, formulas are ok! Language can sometimes be more confusing, and in any case, it is more lengthy. I write this because you tend to explain things with words instead of just writing them down with formulas. (Of course, formulas without explanation are also bad, but you typically spend most words with describing formulas.) Also, you often work with functions without ever giving them a name, and then you need to spend words later to remind the reader which function you mean. Using a symbol will be easier.

For example, I would replace

>  We then claim that $E$ with the restriction $R_E$ of the above realization relation forms a relative exponential of $f$ and $g$ in $\omega{-}Set$. First, it is easy to see that the restriction of $h$ to $E$ is a morphism of $\omega$-sets $(E, R_E) \to (Z, R_Z)$, with $p_1 : N \to N$ working as the partial recursive function in the definition. This is the morphism we will use to consider $(E, R_E)$ as an object of $\omega{-}Set / (Z, R_Z)$.

with (after having named the realization relation $R$)

> Then $(E,R|_E)$ is an $\omega$-set, and $h|_E : (E, R|_E) \to (Z, R_Z)$ is a morphism of $\omega$-sets with realization function $p_1 : N \to N$. We claim that $h|_E$ is the exponential of $f$ and $g$ in $\omega{-}Set / (Z, R_Z)$.

It is much shorter, but tells exactly the same.

(this also fixes an incorrect abuse of notation that does not distinguish a morphism from its domain, which is quite common for slice categories, but not necessary either.)

I know that this is a superficial analysis at this point, but I think it already helps to make the proof easier to understand. I will write more once I actually look at the mathematics.

General tip for longer proofs: Preview and author the math on a temporary content page, which is easier to author and to refresh.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+      Finally, suppose we have a morphism $d : (U, R_U) \to (Z, R_Z)$ and a morphism $j : (U, R_U) \times (Z, R_Z) (X, R_X) \to (Y, R_Y)$ in $\omega{-}\Set / (Z, R_Z)$. Choose partial recursive functions $\varphi_d : \IN \dashrightarrow \IN$ and $\varphi_j : \IN \dashrightarrow \IN$ as in the definition of morphism. Then we can curry $\varphi_j$ into a full recursive function $\varphi_j ' : \IN \to \IN$, such that $\varphi_j '(n) \cdot m \downarrow$ if and only if $\varphi_j(\langle n, m \rangle)$ exists, and in that case $\varphi_j '(n) \cdot m = \varphi_j(\langle n, m \rangle)$. It is now straightforward to check that $n \mapsto \langle \varphi_d(n), \varphi_j'(n) \rangle$ is a partial recursive function, and it shows that the function $\tilde j : U \to [X, Y]_Z$ coming from $[X, Y]_Z$ being a relative exponential in $\Set$ factors through $E$, and it induces a morphism $(U, R_U) \to (E, R_E)$.
```
ScriptRaccoon: `j : (U, R_U) \times (Z, R_Z) (X, R_X) \to (Y, R_Y)`

Here `\times_` is missing

database/data/categories/omega-Set.yaml
ScriptRaccoon: > Hmm... It might be possible to define a "regularly partial morphism classifier"  [...]
 
A more "pedestrian" and "educational" approach:

1. First give a full proof that the category is cartesian closed.
2. Then give a full proof that the category is even locally cartesian closed, and explain which changes are necessary.

Compare this with **Set**. I don't think any beginner in category theory will even be able to understand the proof that **Set** is locally cartesian closed without first having understood why it is cartesian closed (independently from the fact that there are proofs reducing the properties to one another).

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
+id: omega-Set
+name: category of ω-sets
+notation: $\omega{-}\Set$
+objects: pairs $(X, R_X)$ where $X$ is a set and $R_X \subseteq \IN \times X$ is a "realization" relation such that $(p_2)_*(R_X) = X$
+morphisms: 'a morphism $(X, R_X) \to (Y, R_Y)$ is a function $f : X \to Y$ such that there exists a <a href="https://ncatlab.org/nlab/show/partial+recursive+function">partial recursive function</a> $\varphi : \IN \dashrightarrow \IN$ such that whenever we have $n\in \IN, x\in X$ such that $n$ realizes $x$ (i.e. $(n, x) \in R_X$), then $\varphi(n)$ exists and $\varphi(n)$ realizes $f(x)$'
```
ScriptRaccoon: Something like:
 
```suggestion
morphisms: 'a morphism $(X, R_X) \to (Y, R_Y)$ is a function $f : X \to Y$ such that there exists a <a href="https://ncatlab.org/nlab/show/partial+recursive+function">partial recursive function</a> $\varphi : \IN \dashrightarrow \IN$ such that for every $(n,x) \in R_X$, the value $\varphi(n)$ is defined and $(\varphi(n),f(x)) \in R_Y$'
```

See also my comment on formulas. I think the "realization" interpretation can be explained in the description field.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+  Note that in several of the properties below, we will fix a choice of a full recursive pairing function $\langle {-}, {-} \rangle : \IN^2 \to \IN$ with full recursive inverses $p_1, p_2 : \IN \to \IN$. (Standard examples of such a pairing function include $\langle m, n \rangle \coloneqq \binom{m+n+1}{2} + m$, or the function intertwining the binary representations of $m$ and $n$.)
+
+  For a few of the properties, we also fix a choice of <a href="https://en.wikipedia.org/wiki/Admissible_numbering">admissible numbering</a> of the partial recursive functions. (Typical examples of such an admissible pairing include integer encodings of Turing machines; integer encodings of untyped lambda calculus expressions; or integer encodings of the "syntax tree" building up a partial recursive function from the mathematical definition as the smallest class of functions closed under certain operations.) Given such an admissible numbering, we will use the notation that $p \cdot n \downarrow$ if for the $p$th partial recursive function $\varphi_p : \IN \to \IN$, $\varphi_p(n)$ exists, and in this case we will write $p \cdot n$ for its value.
```
ScriptRaccoon:
```suggestion
  For a few of the properties, we also fix a choice of <a href="https://en.wikipedia.org/wiki/Admissible_numbering">admissible numbering</a> of the partial recursive functions. We will use the notation that $p \cdot n \downarrow$ if for the $p$th partial recursive function $\varphi_p : \IN \to \IN$, $\varphi_p(n)$ exists, and in this case we will write $p \cdot n$ for its value.
```

CatDat should not replace or compete with Wikipedia. If the examples are missing in the Wikipedia article, I would add them there. I would only keep the examples here if the proofs depend on them.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+  - property: semi-strongly connected
+    proof: 'Suppose we have two objects $(X, R_X)$ and $(Y, R_Y)$. If $Y$ is non-empty, suppose $y_0 \in Y$, and find $n_0 \in \IN$ such that $n_0$ realizes $y_0$. Then it is easy to see that the constant function with value $y_0$ is a morphism in $\Hom_{\omega{-}\Set}((X, R_X), (Y, R_Y))$, where we can choose $\varphi : \IN \dashrightarrow \IN$ to be the full recursive function with constant value $n_0$. Otherwise, if $Y$ is empty, then the unique function $Y \to X$ is a morphism in $\Hom_{\omega{-}\Set}((Y, R_Y), (X, R_X))$, and in fact any partial recursive function $\IN \dashrightarrow \IN$ works for $\varphi$.'
+
+  - property: well-powered
+    proof: If $(X, R_X) \hookrightarrow (Y, R_Y)$ is a monomorphism, then by the description below, this implies the underlying function $X \to Y$ is injective. Therefore, we can find an equivalent subobject $(X, R_X) \cong (Y', R_{Y'})$ where $Y' \subseteq Y$ by transporting $R_X$ to the image of the underlying function. The collection of possible $Y'$ is then essentially small, and for each $Y'$, the collection of possible $R_{Y'}$ is also essentially small.
```
ScriptRaccoon:
```suggestion
    proof: If $(X, R_X) \hookrightarrow (Y, R_Y)$ is a monomorphism, then by the description below, this implies the underlying function $X \to Y$ is injective. Therefore, we can find an equivalent subobject $(X, R_X) \cong (Y', R_{Y'})$ where $Y' \subseteq Y$ by transporting $R_X$ to the image of the underlying function. The collection of possible $Y'$ is then small, and for each $Y'$, the collection of possible $R_{Y'}$ is also small.
```

The collection of subsets of a set is a set.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+  - property: well-copowered
+    proof: If $(X, R_X) \twoheadrightarrow (Y, R_Y)$ is an epimorphism, then by the description below, this implies the underlying function $X \to Y$ is surjective. Therefore, if $E$ is the kernel pair of this underlying function, then we can find an equivalent quotient object $(X / E, R_{X'})$ by transporting $R_X$ to $X / E$. The collection of possible equivalence relations $E$ is then essentially small, and for each $E$, the collection of possible $R_{X'}$ is also essentially small.
```
ScriptRaccoon:
```suggestion
    proof: If $(X, R_X) \twoheadrightarrow (Y, R_Y)$ is an epimorphism, then by the description below, this implies the underlying function $X \to Y$ is surjective. Therefore, if $E$ is the kernel pair of this underlying function, then we can find an equivalent quotient object $(X / E, R_{X'})$ by transporting $R_X$ to $X / E$. The collection of possible equivalence relations $E$ is then small, and for each $E$, the collection of possible $R_{X'}$ is also small.
```

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+  - property: terminal object
+    proof: 'The $\omega$-set $(1, \IN \times 1)$ is terminal. Namely, for any object $(X, R_X)$, the unique function $! : X \to 1$ is a morphism $(X, R_X) \to (1, \IN \times 1)$, since for example we can take any total recursive function $\varphi : \IN \to \IN$ in the definition.'
+    check_redundancy: false
+
+  - property: generator
```
ScriptRaccoon: Please adjust the order: terminal object, binary products, equalizers (i.e. finite limits), initial object, binary coproducts, coequalizers (i.e. finite colimits), then other things like generator.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
...
+    proof: The terminal object represents the forgetful functor $\omega{-}\Set \to \Set$ which is clearly faithful.
```
ScriptRaccoon: This requires a small proof. For $y \in Y$ we choose $m$ that realizes it, then the constant function $\varphi$ with value $m$ is a realization of the function $y : 1 \to Y$, so that we get a morphism of $\omega$-sets $(1,R \times 1) \to (Y,R_Y)$.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+  - property: cogenerator
+    proof: 'The object $(2, \IN\times 2)$ is a cogenerator. To see this, note that $\Hom_{\omega{-}\Set}({-}, (2, \IN \times 2) \cong \Hom_{\Set}(U{-}, 2)$, where $U : \omega{-}\Set \to \Set$ is the forgetful functor. Therefore, $\Hom_{\omega{-}\Set}({-}, (2, \IN \times 2))$ is isomorphic to the composition of $U$ and the power set functor $P : \Set^{\op} \to \Set$, both of which are faithful.'
```
ScriptRaccoon:
```suggestion
    proof: 'The object $(2, \IN\times 2)$ is a cogenerator. To see this, note that $\Hom_{\omega{-}\Set}({-}, (2, \IN \times 2)) \cong \Hom_{\Set}(U{-}, 2)$, where $U : \omega{-}\Set \to \Set$ is the forgetful functor. Therefore, $\Hom_{\omega{-}\Set}({-}, (2, \IN \times 2))$ is isomorphic to the composition of $U$ and the power set functor $P : \Set^{\op} \to \Set$, both of which are faithful.'
```

Also, can be made displaystyle.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,138 @@
...
+      and the two coprojections are the standard set-level coprojections $i_1 : X \hookrightarrow X \sqcup Y$ and $i_2 : Y \hookrightarrow X \sqcup Y$. First, those coprojections are indeed morphisms of $\omega$-sets, since the functions $n \mapsto 2n$ and $n \mapsto 2n+1$ are full recursive functions $\IN \to \IN$. Also, for any two morphisms $f : (X, R_X) \to (Z, R_Z)$ and $g : (Y, R_Y) \to (Z, R_Z)$, let $\varphi, \psi : \IN \dashrightarrow \IN$ be respective partial recursive functions from the definition. Then there is a unique partial recursive function which sends $2n \mapsto \varphi(n)$ (if the latter exists) and $2n+1 \mapsto \psi(n)$ (if the latter exists), and it shows that $f+g : X\sqcup Y \to Z$ induces a morphism $(X\sqcup Y, R_{X\sqcup Y}) \to (Z, R_Z)$.
+    check_redundancy: false
+
+  - property: coequalizers
+    proof: >-
+      Let $f, g : (X, R_X) \rightrightarrows (Y, R_Y)$ be two parallel morphisms, and let $Q$ be the coequalizer of the underlying functions with projection $p : Y \twoheadrightarrow Q$. Also, let
+      $$R_Q \coloneqq (\id_{\IN} \times p)_*(R_Y) = \{ (n, p(y)) \mid (n, y) \in R_Y \}.$$
```
ScriptRaccoon: You need to write `{\id_{\IN}}` to fix a spacing issue (subtle). Probably I need to fix the definition of the macro`\id`.

database/data/categories/omega-Set.yaml
dschepler: I think it could clean things up to define $X_z$ as $f^*(\{z\})$ with the restricted realization relation, and similarly for $Y_z := g^*(\{z\})$. Then the underlying set of the relative exponential is $\bigsqcup_{z\in Z} Hom(X_z, Y_z)$, and it's natural to define a realizer of an element as a pairing of a realizer of $z$ and a code for the realizer transformer of the morphism. Then the proofs should hopefully look a lot like $z$-parameterized versions of the proofs for cartesian closure that I just added.

database/data/categories/Cat.yaml
```diff
@@ -45,9 +45,6 @@ unsatisfied_properties:
   - property: balanced
     proof: Since we know that <a href="/category/Mon">$\Mon$</a> is not balanced, there is a monoid map $M \to N$ which is a monomorphism and an epimorphism which is not an isomorphism. Then $B(M) \to B(N)$ has the corresponding properties.
 
-  - property: regular
-    proof: See Example 3.14 at the <a href="https://ncatlab.org/nlab/show/regular+category" target="_blank">nLab</a>.
-
```
ScriptRaccoon, line 50: Is this new implication an answer to https://mathoverflow.net/questions/513430 because it makes all these assignments redundant?

I noticed that Example 3.14 is now only used for Top<sub>*</sub> and PMet (and Met).

database/data/category-implications/congruences.yaml
```diff
@@ -126,6 +126,34 @@
...
 
+- id: regular_extensive_consequences
+  assumptions:
+    - extensive
+    - regular
+    - quotients of congruences
+  conclusions:
+    - effective cocongruences
+  proof: >-
+    Suppose $p : X+X' \twoheadrightarrow E$ is a cocongruence, with coreflexivity morphism $r : E \to X$ and cotransitivity morphism $t : E \to E +_X E'$. (Here $X'$ is an isomorphic copy of $X$; $E'$ is an isomorphic copy of $E$; and $E +_X E'$ is the coproduct modulo the relations $p(x') = p(x)'$, i.e. it is equivalent to the pushout of $E \xleftarrow{p\circ i_2} X \xrightarrow{p\circ i_1} E$. Since both these morphisms are (split) monomorphisms, the pushout exists by <a href="/content/pushouts-of-monos-via-congruence-quotients">this result</a>.)
```
ScriptRaccoon, line 137: I find `$E \xleftarrow{p\circ i_2} X \xrightarrow{p\circ i_1} E$` confusing since $i_1,i_2$ mean something different in the diagram below.

Can we maybe state the proof using a pair $p_1, p_2 : X \rightrightarrows E$? (This is my general recommendation for (co)congruences in CatDat. Working with pairs has made several things more clear to me.)

Also expressions like $p(r(e))$ are confusing (used below) since the codomain of $r$ is not the domain of $p$. You actually mean $p(i_1(r(e))$. It is easer to write directly $p_1(r(e))$.

I would then also remove the usage of $X'$ and $E'$. I have suggested this notation before, but I don't think anymore that it is good (at least, in this proof).

If you agree, I would also suggest to rework `regular_epi-regular_extensive_consequences` in the same way.

database/data/category-implications/congruences.yaml
```diff
@@ -126,6 +126,34 @@
...
+    @V p \circ i_1 VV @VV t V @. @V p \circ i_2 VV @VV t V \\
+    E @> i_1 >> E +_X E' @. \quad @. E'  @> i_2 >> E +_X E'.
+    \end{CD}$$
+    For the left diagram, suppose we have generalized elements $e_1, e_2 \in E$ such that $t(e_1) = i_1(e_2) \in E +_X E'$. Then we have two retractions of $t$, $s_1 : E +_X E' \to E$, defined by $e \mapsto p(r(e)), e' \mapsto e'$; and $s_2 : E +_X E' \to E$, defined by $e \mapsto e, e' \mapsto p(r(e)')$. To check for instance that $s_1$ is a retraction, note that $s_1(t(p(x))) = s_1(p(x)) = p(r(p(x)) = p(x)$ and $s_1(t(p(x'))) = s_1(p(x')') = p(x')$ and then use the assumption that $p$ is an epimorphism; the proof for $s_2$ is similar. Therefore, $e_1 = s_1(t(e_1)) = s_1(i_1(e_2)) = p(r(e_2))$. On the other hand, $e_1 = s_2(t(e_1)) = s_2(i_1(e_2)) = e_2$. Therefore, $e_1 = e_2 = p(r(e_2))$, so the generalized element $r(e_2)$ of $X$ maps to $e_1$ and $e_2$ respectively as required. On the other hand, since $p\circ i_1 : X \to E$ is a (split) monomorphism, the uniqueness of the generalized element of $X$ is automatic, completing the proof that the left diagram is a pullback square. The proof for the right diagram is similar.
```
ScriptRaccoon, line 145: It is a bit confusing that you first choose $e_1,e_2$ and then claim that "Then we have two retractions ..." even though their construction is independent of $e_1,e_2$. Isn't this a general lemma about cocongruences that we might extract?

The lemma is especially clear in the dualized version with generalized elements, i.e. w.l.o.g. in **Set**. If $R$ is an equivalence relation on $X$, we have $t : R \times_X R \to R$ mapping $((a,b),(b,c)) \mapsto (a,c)$, and it has two sections $(a,c) \mapsto ((a,a),(a,c))$ and $(a,c) \mapsto ((a,c),(c,c))$.

ScriptRaccoon, line 145: It is confusing to write $e  \mapsto \cdots$ in the definition of $s_1$. You actually mean (and use!) $i_1(e) \mapsto \cdots$.

ScriptRaccoon, line 145: I suggest to use more displaystyle formulas in this paragraph.

database/data/category-implications/congruences.yaml
```diff
@@ -126,6 +126,34 @@
...
+  assumptions:
+    - extensive
+    - regular
+    - quotients of congruences
+  conclusions:
+    - effective cocongruences
+  proof: >-
+    Suppose $p : X+X' \twoheadrightarrow E$ is a cocongruence, with coreflexivity morphism $r : E \to X$ and cotransitivity morphism $t : E \to E +_X E'$. (Here $X'$ is an isomorphic copy of $X$; $E'$ is an isomorphic copy of $E$; and $E +_X E'$ is the coproduct modulo the relations $p(x') = p(x)'$, i.e. it is equivalent to the pushout of $E \xleftarrow{p\circ i_2} X \xrightarrow{p\circ i_1} E$. Since both these morphisms are (split) monomorphisms, the pushout exists by <a href="/content/pushouts-of-monos-via-congruence-quotients">this result</a>.)
+
+    We now claim that we have pullback diagrams
```
ScriptRaccoon, line 139: These two pullback diagrams hold for every cocongruence and can be extracted to a lemma (which I would state for congruences).

database/data/category-implications/congruences.yaml
```diff
@@ -126,6 +126,34 @@
...
+    E @> i_1 >> E +_X E' @. \quad @. E'  @> i_2 >> E +_X E'.
+    \end{CD}$$
+    For the left diagram, suppose we have generalized elements $e_1, e_2 \in E$ such that $t(e_1) = i_1(e_2) \in E +_X E'$. Then we have two retractions of $t$, $s_1 : E +_X E' \to E$, defined by $e \mapsto p(r(e)), e' \mapsto e'$; and $s_2 : E +_X E' \to E$, defined by $e \mapsto e, e' \mapsto p(r(e)')$. To check for instance that $s_1$ is a retraction, note that $s_1(t(p(x))) = s_1(p(x)) = p(r(p(x)) = p(x)$ and $s_1(t(p(x'))) = s_1(p(x')') = p(x')$ and then use the assumption that $p$ is an epimorphism; the proof for $s_2$ is similar. Therefore, $e_1 = s_1(t(e_1)) = s_1(i_1(e_2)) = p(r(e_2))$. On the other hand, $e_1 = s_2(t(e_1)) = s_2(i_1(e_2)) = e_2$. Therefore, $e_1 = e_2 = p(r(e_2))$, so the generalized element $r(e_2)$ of $X$ maps to $e_1$ and $e_2$ respectively as required. On the other hand, since $p\circ i_1 : X \to E$ is a (split) monomorphism, the uniqueness of the generalized element of $X$ is automatic, completing the proof that the left diagram is a pullback square. The proof for the right diagram is similar.
+
+    Now, using extensivity, we can combine these into a pullback diagram
```
ScriptRaccoon, line 147: This is intuitively clear to me, but how can we use the definition of extensivitiy precisely here?

database/data/category-implications/congruences.yaml
```diff
@@ -126,6 +126,34 @@
...
+    $$\begin{CD}
+    X+X' @> p >> E \\
+    @VVV @VV t V \\
+    E+E' @>>> E +_X E'.
+    \end{CD}$$
+    Since both morphisms $X \to E$ and $X \to E'$ used in the pushforward are monomorphisms, by the construction <a href="/content/pushouts-of-monos-via-congruence-quotients">here</a>, the morphism $E+E' \to E +_X E'$ is a regular epimorphism. Therefore, by the assumption that the category is regular, we have that $p$ is in fact a regular epimorphism.
```
ScriptRaccoon, line 153:
```suggestion
    Since both morphisms $X \to E$ and $X \to E'$ used in the pushout are monomorphisms, by the construction <a href="/content/pushouts-of-monos-via-congruence-quotients">here</a>, the morphism $E+E' \to E +_X E'$ is a regular epimorphism. Therefore, by the assumption that the category is regular, we have that $p$ is in fact a regular epimorphism.
```

database/data/category-implications/congruences.yaml
```diff
@@ -126,6 +126,34 @@
...
+
+    From here, essentially the same argument as in <a href="/category-implication/regular_epi-regular_extensive_consequences">this proof</a> applies to show that $E$ is effective.
```
ScriptRaccoon, line 155:
```suggestion
    From here, <a href="/category-implication/regular_epi-regular_extensive_consequences">this proof</a> shows that $E$ is effective. This proof applies since the assumption that the category is epi-regular has only been used to make sure that $p$ is a regular epimorphism.
```

Right?

database/data/categories/LRS_R.yaml
```diff
@@ -88,9 +88,6 @@ unsatisfied_properties:
   - property: cartesian filtered colimits
     proof: This is Corollary 4(b) <a href="/content/Top-embeds-in-LRS">here</a>.
 
-  - property: regular
-    proof: This is Corollary 4(c) <a href="/content/Top-embeds-in-LRS">here</a>.
-
```
ScriptRaccoon, line 93: Do we want to remove it then from the content page?

database/data/categories/Meas.yaml
ScriptRaccoon, line 1: Quite amazing that these non-trivial proofs for **Bin** and **Meas** can be removed.

shared/structure.history.json
ScriptRaccoon, line 1: Can you please rebase on `main`? It also contains the changes to `pnpm-workspace.yaml`.

ScriptRaccoon:
> Currently unresolved properties  [...]

With all of the categories that I have added in the last weeks, I have decided every single property. This was a lot of work, but Gemini Pro has helped me a lot to find the proofs (I can recommend using it), and I think the database has the best value when there are no questions left. But, of course, if something just does not work, leave it open and/or write an issue.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,158 @@
...
+
+      From this, we see that there can be no $\aleph_0$-indexed power of $N$. Indeed, if $X$ were such a power, then for any sequence $(a_n) \in \IN^{\IN}$, we would have a corresponding sequence of morphisms $1 \to \IN$, inducing a morphism $1 \to X$ and thus an element $x$ of the underlying set. The corresponding function $g : X \to \IN^{\IN}$ would then have to have value $g(x) = (a_n)$. Thus, $g$ would be surjective, contradicting the previous paragraph.
+
+special_objects:
+  initial object:
+    description: $(\varnothing, \varnothing)$
+  terminal object:
+    description: $(1, \IN \times 1)$
+  coproducts:
+    description: The binary coproduct of $(X, R_X)$ and $(Y, R_Y)$ is $X \sqcup Y$ with realization relation $$R_{X\sqcup Y} \coloneqq \{ (2n, x) \mid (n, x) \in R_X \} \cup \{ (2n+1, y) \mid (n, y) \in R_Y \}.$$
```
ScriptRaccoon, line 139: The entry is for all coproducts. If the description is only for finite or binary coproducts, please indicate this just like on the page for, say, [FinSet](https://catdat.app/category/FinSet).

Same remark applies to products.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,158 @@
...
+  regular monomorphisms:
+    description: 'A morphism $f : (X, R_X) \to (Y, R_Y)$ is a regular monomorphism if and only if the underlying function $X \to Y$ is injective, and there exists a partial recursive function $\psi : \IN \dashrightarrow \IN$ such that whenever we have $n\in \IN$ and $x\in X$ such that $n$ realizes $f(x) \in Y$, then $\psi(n)$ exists and $\psi(n)$ realizes $x\in X$'
+    proof: 'Following the constructions above, we can see that for any morphism $f : (X, R_X) \to (Y, R_Y)$, the equalizer of the cokernel pair of $f$ is $(\im(f), R_Y \cap (\IN \times \im(f)))$; and the condition above is equivalent to the condition that the canonical map $X \to \im(f)$ induces an isomorphism of $\omega$-sets $(X, R_X) \to (\im(f), R_Y \cap (\IN \times \im(f)))$.'
+  epimorphisms:
+    description: Morphisms of $\omega$-sets whose underlying set function is surjective
+    proof: Sufficiency of this condition is easy to show. For necessity, use the fact that $\Hom_{\Set}(X, Z) \cong \Hom_{\omega{-}\Set}((X, E_X), (Z, \IN \times Z))$ to show that if $f$ is an epimorphism in $\omega{-}\Set$, then the underlying set function of $f$ is an epimorphism in $\Set$.
```
ScriptRaccoon, line 155:
```suggestion
    proof: Sufficiency of this condition is trivial. For necessity, note that $\Hom_{\Set}(X, Z) \cong \Hom_{\omega{-}\Set}((X, E_X), (Z, \IN \times Z))$. Hence, the forgetful functor is a left adjoint, and therefore preserves epimorphisms.
```

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,158 @@
...
+    description: $(\varnothing, \varnothing)$
+  terminal object:
+    description: $(1, \IN \times 1)$
+  coproducts:
+    description: The binary coproduct of $(X, R_X)$ and $(Y, R_Y)$ is $X \sqcup Y$ with realization relation $$R_{X\sqcup Y} \coloneqq \{ (2n, x) \mid (n, x) \in R_X \} \cup \{ (2n+1, y) \mid (n, y) \in R_Y \}.$$
+  products:
+    description: The binary product of $(X, R_X)$ and $(Y, R_Y)$ is $X \times Y$ with realization relation $$R_{X\times Y} \coloneqq \{ (\langle n, m \rangle, (x, y)) \mid (n, x) \in R_X \land (m, y) \in R_Y \}.$$
+
+special_morphisms:
+  isomorphisms:
```
ScriptRaccoon, line 144: Can you add a comment to the file (in the comment section, see `Sh(X).yaml` for an example) that gives an example of a mono- and epimorphism that is not an isomorphism? Or, maybe even better, make a (redundant) assignment that the category is not balanced?

I ask because this would explain why this description of isomorphisms is relevant.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,158 @@
+id: omega-Set
+name: category of ω-sets
+notation: $\omega{-}\Set$
+objects: pairs $(X, R_X)$ where $X$ is a set and $R_X \subseteq \IN \times X$ is a "realization" relation such that $(p_2)_*(R_X) = X$
+morphisms: 'a morphism $(X, R_X) \to (Y, R_Y)$ is a function $f : X \to Y$ such that there exists a <i>realizer transformer compatible with $f$</i>: this is a <a href="https://ncatlab.org/nlab/show/partial+recursive+function">partial recursive function</a> $\varphi : \IN \dashrightarrow \IN$ such that for every $(n,x) \in R_X$, the value $\varphi(n)$ is defined and $(\varphi(n),f(x)) \in R_Y$'
+description: >-
```
ScriptRaccoon, line 6: Can we give a description of this category that is accessible for readers without looking at the effective topos? Because this was the initial goal: make the construction of the effective topos easier to follow. So the references should (mostly) go only in one direction.

For example: what is an intuitive picture for the pair $(X,R_X)$? Why do we say that $n$ realizes $x$ when $(n,x) \in R_X$? This seems to be hinting at some interpretation, but this interpretation is never mentioned. Maybe the word "assembly" is also relevant.

I don't have answers to these questions. I still try to understand the basics here.

Maybe this information can also be added to the linked nLab article.

database/data/categories/omega-Set.yaml
```diff
@@ -0,0 +1,158 @@
...
+
+  - property: well-copowered
+    proof: If $(X, R_X) \twoheadrightarrow (Y, R_Y)$ is an epimorphism, then by the description below, this implies the underlying function $X \to Y$ is surjective. Therefore, if $E$ is the kernel pair of this underlying function, then we can find an equivalent quotient object $(X / E, R_{X / E})$ by transporting $R_X$ to $X / E$. The collection of possible equivalence relations $E$ is then small, and for each $E$, the collection of possible $R_{X / E}$ is also small.
+
+  - property: terminal object
+    proof: 'The $\omega$-set $(1, \IN \times 1)$ is terminal. Namely, for any object $(X, R_X)$, the unique function $! : X \to 1$ is a morphism $(X, R_X) \to (1, \IN \times 1)$, since for example any total recursive function $\varphi : \IN \to \IN$ forms a compatible realizer transformer.'
```
ScriptRaccoon, line 35: This can also be explained by using the fact (mentioned in the proof for epis, but maybe one can make this more prominent) that $X \mapsto (X, N \times X)$ is right adjoint to the forgetful functor, and right adjoints preserve terminal objects.

database/data/categories/Cat.yaml
dschepler: I guess it's a partial answer, given that it only works for the extensive categories (with coequalizers) where you can find a non-effective cocongruence. If I remember correctly, I traced through the proof in the case of **Top** and it gave this pullback of a regular epimorphism which isn't a regular epimorphism:
$$\begin{CD}
\{ a, d \}_d @>>> \{ a, d \}_i \\
@VVV @VVV \\
\{ a, b \}_i \sqcup \{ c, d \}_i @>>> \{ a, b = c, d \}_i
\end{CD}$$

database/data/category-implications/congruences.yaml
dschepler: If you have morphisms $U \to E$ and $U \to E + E'$ with equal compositions to $E +_X E'$, then use extensionality to split the second into $U_1 \to E$ and $U_2 \to E'$ with $U \cong U_1 + U_2$; then the individual pullbacks induce morphisms $U_1 \to X$ and $U_2 \to X'$, and their sum gives the required morphism $U \to X + X'$.

database/data/categories/LRS_R.yaml
dschepler: I guess we could also keep it in with `check_redundancy: false`

database/data/categories/omega-Set.yaml
dschepler: I also haven't found motivation for the definitions in any of the sources I've found, except that in the end it gives a topos with desirable properties of the internal logic.

Though I suppose a possible inkling of an answer in the form an example could be: You could fix an enumeration of the intervals in $\mathbb{R}$ with rational endpoints, and then define an $\omega$-set whose underlying set is $\mathbb{R}$ and whose realization relation is that $n$ realizes $x$ exactly when $x$ is in the $n$th interval. Then for $f : \mathbb{R} \to \mathbb{R}$, a compatible realizer transformer would correspond to a sound approximation of $f$ in terms of "interval arithmetic".

Though that example also has its shortcomings. Namely, in interval arithmetic, you usually also care whether the approximation more or less has some reasonable "tightness" properties, whereas the category of $\omega$-sets doesn't care in the least about tightness for that example, only about soundness. Also, the morphisms do not contain any particular realizer transformer in their data, only the existence of at least one is required. (So, I think most "naturally occurring" locally bounded functions $\mathbb{R} \to \mathbb{R}$ will be morphisms in that category, except for pathological examples such as applying the Busy Beaver function to the floor of $x$.)

So very roughly, I've sort of been thinking of an $\omega$-set as an abstract set but where there are some "pieces of information" about elements of the set which have some "computational relevance".

database/data/categories/omega-Set.yaml
dschepler: I was also thinking that I could eventually add both functors to the database, along with some corresponding functors for the effective topos (the right adjoint is conventionally called $\nabla$, though that's also the name of a similar right adjoint for the effective category, so I would probably call them $\nabla_{\omega{-}Set}$ and $\nabla_{Eff}$ respectively), and also a couple functors between $\omega{-}Set$ and $Eff$.

database/data/categories/omega-Set.yaml
dschepler: Maybe a more illustrative example would be the set of computable reals: there, for example, we could define a realizer to be $\langle p, R \rangle$ where $p$ is a code for a recursive function $f : \mathbb{Z} \to \mathbb{Z}$ and $R$ is a uniform bound on $|f(m+n) - f(m) - f(n)|$, such that $x$ is the limit of $f(n)$ as $n \to \pm \infty$. (So we think of $f(n)$ as being an approximation of $n\alpha$ for some computable real $\alpha$.)

dschepler:
> With all of the categories that I have added in the last weeks, I have decided every single property. This was a lot of work, but Gemini Pro has helped me a lot to find the proofs (I can recommend using it), and I think the database has the best value when there are no questions left. But, of course, if something just does not work, leave it open and/or write an issue.

OK, I'm stuck on the current set of unresolved properties:
is [accessible](http://localhost:5173/category-property/accessible)
is [coaccessible](http://localhost:5173/category-property/coaccessible)
has an [extremal generator](http://localhost:5173/category-property/extremal_generator)
is [ℵ₁-accessible](http://localhost:5173/category-property/aleph1-accessible)
has [ℵ₁-cofiltered limits](http://localhost:5173/category-property/aleph1-cofiltered_limits)
has [ℵ₁-filtered colimits](http://localhost:5173/category-property/aleph1-filtered_colimits)

I don't have Gemini Pro at the moment; and when I asked free Google AI on some of these questions, it just ended up in the sort of situation where it was just spouting nonsense, and when corrected it just came up with some other nonsense to spout. On a lot of them, it ended up referring again to the van Oosten paper which only proves that a certain set of objects doesn't form a dense subcategory of the effective topos - nothing I could find in that paper that would give direct answers to any of these questions, or even the corresponding questions in the effective topos.

The closest I've come is a pathological-looking ℵ₁-filtered diagram inspired by the van Oosten paper: let $(s_\alpha)_{\alpha < \omega_1}$ be an $\omega_1$-Aronszajn sequence, which is a sequence of injective functions $\alpha \to \omega_1$ such that whenever $\alpha < \beta$, then $s_\beta |_{\alpha}$ and $s_\alpha$ differ only at finitely many inputs. Now define $X_\alpha$ to be the $\omega$-set with underlying set $\alpha$ and with realizability relation that $n$ realizes $\gamma < \alpha$ if and only if $s_\alpha(\gamma) = n$. Then the almost-compatibility condition implies that only finitely many realizers have to be "reassigned", which is certainly possible to do with a full recursive function; therefore, the inclusion map $\alpha \hookrightarrow \beta$ induces a morphism $X_\alpha \to X_\beta$. Any colimit of this diagram would have to have underlying set bijective with $\omega_1$ (and the cocone maps are equivalent to the inclusions) since the forgetful functor is a left adjoint. But beyond that, I haven't found a proof of a contradiction, nor a construction of a realization relation on $\omega_1$ which makes it a colimit. (Other than the vague idea of using the extremal cogenerator of $P(\mathbb{N}) \setminus \{\emptyset\}$ to maybe extract some information on what the realizers would have to look like.)

Otherwise, I'm still working on revising the proofs of local cartesian closure and having effective cocongruences, and maybe trying to find a better overview for the category description.

database/data/categories/LRS_R.yaml
ScriptRaccoon: No let's remove it as you did.

ScriptRaccoon:
What is the status of the comments I made but that are not resolved yet? Are you planning to work on these? I am basically waiting for that before I continue the review.

database/data/category-implications/congruences.yaml
ScriptRaccoon: I suggest to add this to the proof then.

database/data/categories/omega-Set.yaml
ScriptRaccoon: Thank you. How can we make sure that these remarks are not buried in this PR? As before, I suggest adding them to the nLab page (https://ncatlab.org/nlab/show/assembly perhaps).

ScriptRaccoon:
Response to https://github.com/ScriptRaccoon/CatDat/pull/373#issuecomment-5801726136:

> I don't have Gemini Pro at the moment;

I also don't pay for Google Gemini, but you get some free credits per day. I have a couple of Google accounts, so that I have even more free credits.

> and when I asked free Google AI on some of these questions, it just ended up  [...]

Oh yeah, even Gemini Pro also produced a lot of bullshit in the last weeks. Maybe they changed something, but it is also possible that our questions get more complex ... And Google Gemini with the free Flash model is pretty much useless for (advanced) mathematics.

> The closest I've come is a pathological-looking ℵ₁-filtered diagram [...]
 
I see. We can merge this PR without deciding $\aleph_1$-filtered colimits (from my side), but I wonder how we can save this potential counterexample in the repo, so that we or someone else (!) can perhaps later complete it; and find an approach for partial results that also works in future. Maybe just a text file named like the category?
 
> Otherwise, I'm still working on revising the proofs of local cartesian closure and having effective cocongruences, and maybe trying to find a better overview for the category description.
 
Great!

dschepler:
FYI, I just posted https://math.stackexchange.com/questions/5150285/what-is-the-motivation-for-the-definition-of-omega-sets to ask about how to motivate the definition or explain how to think about "realizers".

